PR #1, opened 2020-03-12
Tickets/dm 23716

jgates108:


python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+        out_str = process.communicate()
+        process.wait()
+        if process.returncode != 0:
+            print("out=", out_str)
+        return process.returncode, out_str
+
+    def _readDatagenConfig(self):
+        """ Create a Chunker and spec using the same configuration file as the datagen.py.
+        """
+        spec_globals = dict()
```
ctslater: I think `spec_globals = {}` is more common

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+                A list chunk ids that need to be generated edgeOnly.
+
+        Note
+        ----
+            To generate the overlap file, there needs to be the
+            complete csv file for chunk_id and either complete
+            or edge only csv files for all of its neighbors.
+        """
+        success = True
+        foundCsv = list()
```
ctslater: I think `foundCsv = []` is more common.

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+                if len(chunkRecvSet) == 0:
+                    # no more chunks, close the connection
+                    print("No more chunks to create, exiting")
+                    self._loop = False
+                else:
+                    createdChunks = list()
+                    withOverlapChunks = list()
+                    haveAllCsvChunks = list()
+                    ingestedChunks = list()
+                    for chunk in chunkRecvSet:
```
ctslater: This loop and the loop at line 716 feel like a separate function. It's a little hard to tell what are their outputs vs. their intermediate variables.

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+                    if len(haveAllCsvChunks):
+                        self._startTransaction()
+                    try:
+                        for chunk in haveAllCsvChunks:
+                            if self._createOverlapTables(chunk):
+                                print("created overlap for chunk", chunk)
+                                withOverlapChunks.append(chunk)
+                            else:
+                                print("ERROR failed to create overlap for chunk", chunk)
+                    except Exception as exc:
```
ctslater: If you can put a specific exception here to catch, it will avoid accidentally hiding things like exceptions due to syntax errors.

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+                            print("Created all needed edgeOnly for ", nCh)
+                            # Put hardlinks to all the files needed for a chunk in
+                            # a specific directory for the partioner to use to create
+                            # the overlap tables and so on.
+                            if self._fillChunkDir(chunk, neighborChunks):
+                                haveAllCsvChunks.append(chunk)
+                    # Generate overlap tables and files for ingest.
+                    # Start the transaction
+                    abort = False
+                    if len(haveAllCsvChunks):
```
ctslater: Not exactly sure what the control flow is supposed to be if haveAllCsvChunks is empty; should this have something like `else: continue`? (That is, I can't easily tell if any the remainder of this loop is important to do when haveAllCsvChunks is empty)

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+            # Write those files to the partitioner config directory
+            pCfgDir = os.path.join(self._target_dir, self._pt_cfg_dir)
+            for it in pCfgDict.items():
+                pCfgName = os.path.join(pCfgDir, it[1][0])
+                print("writing ", it[0], "name=", pCfgName)
+                with open(pCfgName, "w") as fw:
+                    fw.write(it[1][1])
+
+            # Start creating and ingesting chunks.
+            while self._loop:
```
ctslater: Doesn't need the `self`

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+                self._client.clientReqChunks(self._chunksPerReq)
+                chunkListRecv, problem = self._client.clientRecvChunks()
+                if problem:
+                    print("WARN there was a problem with", chunkListRecv)
+                chunkRecvSet = set(chunkListRecv)
+                if len(chunkRecvSet) == 0:
+                    # no more chunks, close the connection
+                    print("No more chunks to create, exiting")
+                    self._loop = False
```
ctslater: If you `break` here, I think you can get rid of the giant `else:` clause.

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+                    if len(ingestedChunks) == 0:
+                        print("ERROR no chunks were successfully ingested, ending program")
+                        self._loop = False
+
+                    # Client sends the list of completed chunks back
+                    while True:
+                        # Keep sending created chunks back until there are none left.
+                        # If ingestedChunks is empty initially, the server needs to be sent an
+                        # empty list to indicate the failure to ingest anything.
+                        ingestedChunks = self._client.clientReportChunksComplete(ingestedChunks)
```
ctslater: Overwriting a list of work with a return value of work _not_ successfully performed is a kind of confusing pattern. If you kept a list of the chunks successfully reported, a few set operations would let you find the work remaining pretty easily.

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+                    success, status, content = self._endTransaction(abort)
+                    if success and not abort:
+                        for chunk in withOverlapChunks:
+                            ingestedChunks.append(chunk)
+                    else:
+                        print("Failed ingest, transaction failed or aborted ", status, content)
+                    # If no chunks were created, likely fatal error. Asking for more
+                    # chunks to create would just cause more problems.
+                    if len(ingestedChunks) == 0:
+                        print("ERROR no chunks were successfully ingested, ending program")
```
ctslater: `sys.exit()`?

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+            # Delete files for this chunk if they exist.
+            if not self.removeFilesForChunk( chunk_id, edge_only=True, complete=True):
+                print("WARN failed to remove all files for chunk=", chunk_id)
+        # Genrate the chunk parquet files.
+        options = " "
+        if edge_only:
+            options += " --edgeonly "
+        cmdStr = ("python " + self._datagenpy + options +
+            " --chunk " + str(chunk_id) + " " + self._gen_arg_str + " " + self._cfg_file_name)
+        genResult, genOut = self.runProcess(cmdStr)
```
ctslater, line 439: It should be possible to call the DataGenerator class directly; I don't see the advantage of running it in a subprocess.

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+        success, status, content = self._ingest.endTransaction(t_id, abort)
+        return success, status, content
+
+    def run(self):
+        """Connect to the server and do everything until the server
+        runs out of chunks for this client to generate and ingest.
+        """
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.connect((self._host, self._port))
+            self._client = DataGenConnection(s)
```
ctslater: When I see `self._client` later in this function, I get confused because this _is_ the DataGenClient class. Maybe `self._conn` ?

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+            while not pCfgName == "":
+                self._client.clientReqPartitionCfgFile(pCfgIndex)
+                indx, pCfgName, pCfgContents = self._client.clientRespPartionCfgFile()
+                if indx != pCfgIndex:
+                    self.success = False
+                    raise RuntimeError("Client got wrong pCfgIndex=", pCfgIndex,
+                                       "indx=", indx, pCfgName)
+                print("pCfgName=", pCfgName)
+                if not pCfgName == "":
+                    pCfgDict[pCfgIndex] = (pCfgName, pCfgContents)
```
ctslater, line 721: Is there a reason not to use `pCfgName` as the dictionary key, instead of storing a tuple?

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+                pCfgIndex += 1
+            self._pt_cfg_dict = pCfgDict
+            # Write those files to the partitioner config directory
+            pCfgDir = os.path.join(self._target_dir, self._pt_cfg_dir)
+            for it in pCfgDict.items():
```
ctslater: If you use `pCfgName` as the dictionary key, then this can be
```
for pCfgName, pCfgContents in pCfgDict.items():
```

ctslater (commented):
Since there's a lot of code to look at here (which is overall very impressive), I'm going to try to do one-file-at-a-time and send over comments after each one. This is just covering DataGenClient.py.

bin/datagen.py
```diff
@@ -26,8 +36,15 @@
...
+    tables = dataGen.make_chunk(chunk_id, num_rows=row_counts, seed=seed,
+                                edge_width=edge_width, edge_only=edge_only)
+
+    print("tables=", tables)
 
     for table_name, table in tables.items():
-        table.to_parquet("chunk{:d}_{:s}.parquet".format(chunk_id, table_name))
+        edgeType = "CT"  # complete
+        if edge_only: edgeType = "EO" # edge only
+        table.to_parquet("chunk{:d}_{:s}_{:s}.parquet".format(chunk_id, edgeType, table_name))
```
ctslater: If you change this to `.to_csv()`, then there's no need to invoke a separate conversion step. I tried this for an example chunk and the results look the same to me (specifying `header=False` and `index=False`).

bin/edgefirsttst.py
```diff
@@ -0,0 +1,108 @@
+
+import argparse
+import pandas as pd
```
kennywlo: Unused import statement for pandas.

bin/datagen.py
```diff
@@ -26,8 +36,15 @@
     if("ForcedSource" in spec):
         row_counts["ForcedSource"] = None
 
-    tables = dataGen.make_chunk(chunk_id, num_rows=row_counts)
+    seed = 1
+    edge_width = 0.017 # degrees
```
kennywlo: Should seed and edge_width be configurable, with defaults, from a file?

bin/datagen.py
kennywlo: Concur here.  The csv support in pandas, with nullable integer dtypes, makes it quite pleasant.

bin/edgefirsttst.py
```diff
@@ -0,0 +1,108 @@
...
+def edgeOnlyContainedInComplete(chunk_id, objectCount, visitCount, spec):
+    """ Check if all the edge only rows can be matched with identical rows
+    in the complete table table.
+    """
+
+    success = None
+    dataGen = DataGenerator(spec)
+
+    seed = 1
+    edge_width = 0.017 # degrees
```
kennywlo: Again, should seed and edge_width be configurable, e.g. via a file?

bin/edgefirsttst.py
```diff
@@ -0,0 +1,108 @@
+
+import argparse
+import pandas as pd
+from lsst.dax.data_generator import columns
+from lsst.dax.data_generator import DataGenerator
+
+
+def edgeOnlyContainedInComplete(chunk_id, objectCount, visitCount, spec):
```
kennywlo: Argument names should be lowercase.

bin/edgefirsttst.py
```diff
@@ -0,0 +1,108 @@
...
+                        break
+                if match:
+                    found = True
+                    break
+            if not found:
+                print("Failed for table=", tblNameEO, " to find rowEO=", rowEO, " in complete table")
+                success = False
+                return success
+        print("rowsChecked=", rowsChecked)
+
```
kennywlo: The syntax of this quadruple-loop construct can be improved: 1) variable names to be lowercase.  2) that long print stmt can be shortened using f-string.  3) return False, instead of success 4) a general comment on what this loop aims to accomplish would be helpful.

bin/edgefirsttst.py
```diff
@@ -0,0 +1,108 @@
...
+        spec_globals = {}
+        exec(f.read(), spec_globals)
+        assert 'spec' in spec_globals, "Specification file must define a variable 'spec'."
+        spec = spec_globals['spec']
+
+    if not edgeOnlyContainedInComplete(args.chunk, args.objects, args.visits, spec):
+        success = False
+
+    if success is None:
+        success = True
```
kennywlo: Do we need to distinguish the special case success being None?  A helpful print message?

python/lsst/dax/data_generator/generator.py
```diff
@@ -1,23 +1,57 @@
...
+    def __init__(self, colNames, generator, position):
+        self.colNames = colNames
+        self.position = position
+        self.generator = generator
+        self.block = None
+
+    def __repr__(self):
+        return ("{colNames:" + self.colNames + ' position:' + str(self.position)
+              + " generator:" + str(self.generator) + ' block:' + str(self.block))
+
```
kennywlo, line 42: Need another blank line per PEP8.

python/lsst/dax/data_generator/generator.py
```diff
@@ -100,35 +140,65 @@ def make_chunk(self, chunk_id, num_rows=None):
         else:
             rows_per_table = defaultdict(lambda: num_rows)
 
-        for table in self._resolve_table_order(self.spec):
-            column_generators = self.spec[table]["columns"]
+        resolvedOrder = self._resolve_table_order(self.spec)
+        tableColumns = dict()
+        newRowsPerTable = dict()
+
```
kennywlo, line 145: Local variables to be lowercase.

python/lsst/dax/data_generator/generator.py
```diff
@@ -100,35 +140,65 @@ def make_chunk(self, chunk_id, num_rows=None):
...
-                        individual_obj = column_generator(
-                            chunk_id, rows_per_table[table],
-                            prereq_row=output_tables[prereq_rows].iloc[n],
+                        preq_rw = output_tables[prereq_rows].iloc[n]
+                        colInfo.block = colInfo.generator(
+                            chunk_id, rows_per_table[table], seed,
+                            prereq_row=preq_rw,
                             prereq_tables=prereq_table_contents)
-                        self._add_to_list(individual_obj, output_columns, split_column_names)
+                        self._add_to_list(colInfo.block, output_columns, split_column_names)
```
kennywlo: For the nested for loops, any meaningful progress message can be printed along the way?  Say, in the format of "X out of Y being processed..."

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+        self._pt_cfg_dict = None # Dictionary that stores partioner config files.
+        self.makeDir(self._target_dir)
+        self.makeDir(os.path.join(self._target_dir, self._pt_cfg_dir))
+
+        self._ingest = None
+        self._skip_ingest = True
+        self._db_name = ''
+        self._transaction_id = -1
+
+    def _setIngest(self, ingest_dict):
```
kennywlo, line 83: Function should be lowercase.

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+        Note
+        ----
+        The keys in ingest_dict should match those in servRespInit and clientRespInit.
+        """
+        ingd = ingest_dict
+        self._ingest = DataIngest(ingd['host'], ingd['port'], ingd['user'], ingd['auth'])
+        self._skip_ingest = ingd['skip']
+        self._db_name = ingd['db']
+
+    def createFileName(self, chunk_id, table_name, ext, edge_only=False, use_targ_path=False):
```
kennywlo, line 105: Function should be lowercase.

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+        typeStr = 'CT_'
+        if edge_only: typeStr = 'EO_'
+        # If the tabelName is a wildcard, don't use typeStr
+        if table_name == '*': typeStr = ''
+        fn = 'chunk' + str(chunk_id) + '_' + typeStr + table_name + '.' + ext
+        if use_targ_path:
+            fn = os.path.join(self._target_dir, fn)
+        return fn
+
+    def makeDir(self, dir_name):
```
kennywlo, line 137: Function should be lowercase.

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+        """
+        try:
+            os.mkdir(dir_name)
+        except OSError as err:
+            if err.errno != errno.EEXIST:
+                print("ERROR directory creation", dir_name, err)
+                return False
+        return True
+
+    def removeFile(self, f_name):
```
kennywlo, line 158: Function should be lowercase.

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,786 @@
...
+        """
+        print("removing file", f_name)
+        try:
+            os.remove(f_name)
+        except OSError as err:
+            print("ERROR remove failed", f_name, err)
+            return False
+        return True
+
+    def runProcess(self, cmd, cwd=None):
```
kennywlo, line 169: Function should be lowercase.

python/lsst/dax/distribution/DataGenConnection.py
```diff
@@ -0,0 +1,727 @@
...
+        sep = self.COMPLEXSEP
+        print("ingest_dict=", ingest_dict)
+        skip_val = '0'
+        if ingest_dict['skip']:
+            skip_val = '1'
+        msg = (name + sep + arg_string + sep + cfg_file_contents
+            + sep + ingest_dict['host'] + sep + str(ingest_dict['port'])
+            + sep + ingest_dict['user'] + sep + ingest_dict['auth']
+            + sep + ingest_dict['db'] + sep + skip_val)
+        self._send_msg(self.S_INIT_R, msg)
```
kennywlo, line 253: Need blank line.

python/lsst/dax/distribution/DataGenConnection.py
```diff
@@ -0,0 +1,727 @@
...
+        Parameters
+        ----------
+        index : int
+            Indicates which partitioner configuration file the client wants.
+            The client starts at 0 and keeps incrmenting by 1 until the server
+            indicates the file does not exist. The number of files varies
+            depending on the database.
+        """
+        print("clientReqChunks C_PCFGR")
+        self._send_msg(self.C_PCFG_R, str(index))
```
kennywlo, line 296: Need blank line.

python/lsst/dax/distribution/DataGenConnection.py
```diff
@@ -0,0 +1,727 @@
...
+        ------
+        index : int
+            Indicates which partitioner configuration file the client wants.
+        """
+        print("servRespPartitionCfgFile C_PCFGR")
+        msg_id, msg, msg_len = self._recv_msg()
+        if not msg_id == self.C_PCFG_R:
+            self.warnings += 1
+            raise DataGenError('ERROR servRespPartitionCfgFile' + str(msg_id) + ' ' + str(msg_len) + ' ' + msg)
+        return int(msg)
```
kennywlo, line 311: Need blank line.

python/lsst/dax/distribution/DataGenConnection.py
```diff
@@ -0,0 +1,727 @@
...
+        file_name : str
+            The name of the file at the index. This should be an empty
+            string if there is no file at that index.
+        file_contents : str
+            The contents of the configuration file. This may be empty.
+        """
+        print("servSendPartionCfgFile S_PCFG_A", index, file_name, len(file_contents))
+        sep = self.COMPLEXSEP
+        msg = str(index) + sep + file_name + sep + file_contents
+        self._send_msg(self.S_PCFG_A, msg)
```
kennywlo, line 330: Need blank line.

python/lsst/dax/distribution/DataGenServer.py
```diff
@@ -0,0 +1,513 @@
...
+import sys
+import threading
+import yaml
+
+from enum import Enum
+
+from DataGenConnection import DataGenConnection
+from DataGenConnection import DataGenError
+from DataIngest import DataIngest
+
```
kennywlo, line 36: Need another blank line.

python/lsst/dax/distribution/DataGenServer.py
```diff
@@ -0,0 +1,513 @@
...
+    client_id : int
+        The id of the client program generating the chunk.
+    client_addr : string
+        The IP address of the client generating the chunk.
+    """
+
+    def __init__(self, chunk_id):
+        self.chunk_id = chunk_id
+        self.gen_stage = GenerationStage(GenerationStage.UNASSIGNED)
+        self.client_id = '-1'
```
kennywlo, line 73: Should be -1 as integer?

python/lsst/dax/distribution/DataGenServer.py
```diff
@@ -0,0 +1,513 @@
...
+        self.client_addr = None
+
```
kennywlo, line 75: client_addr string type - use "" instead?

python/lsst/dax/distribution/DataGenServer.py
```diff
@@ -0,0 +1,513 @@
...
+        self._servAccept()
+        print("Done, generated ", self._total_generated_chunks)
+        print("chunks failed chunks:", self.chunksInState([GenerationStage.LIMBO, GenerationStage.ASSIGNED]))
+        counts = {GenerationStage.UNASSIGNED:0,
+            GenerationStage.ASSIGNED:0,
+            GenerationStage.FINISHED:0,
+            GenerationStage.LIMBO:0}
+        for chk in self._chunks_to_send:
+            chk_info = self._chunks_to_send[chk]
+            counts[chk_info.genStage] += 1
```
kennywlo: Is genStage attribute defined?

python/lsst/dax/distribution/DataIngest.py
```diff
@@ -0,0 +1,343 @@
...
+        if self._id > -1:
+            success, status, content = self._data_ingest.endTransaction(self._id, abort=False)
+        if not success:
+            print("ERROR Transaction end failed ", self._db_name, self._id, status, content)
+            raise RuntimeError('Transaction failed ' + self._db_name + " trans_id="+ str(self._id)
+                                + " status=" + str(status) + " content=" + str(content))
+        return True
+
+
+
```
kennywlo: Too many blank lines.

python/lsst/dax/distribution/DataIngest.py
```diff
@@ -0,0 +1,343 @@
...
+        ----
+        This calls the external program 'qserv-replica-file-ingest' to
+        actually send the file.
+        """
+        cmd = ('qserv-replica-file-ingest FILE ' + host + ' ' + str(port) + ' '
+            + str(transaction_id) + ' ' + table + ' P ' + f_path + ' --verbose --columns-separator=TAB')
+        print("cmd=", cmd)
+        process = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+        out_str = process.communicate()
+        process.wait()
```
kennywlo: Why we need to use subprocess() for a synchronous call?

kennywlo (commented):
General issues with syntax in regard to PEP8, e.g. variable naming.  The _requirements.txt_ still needs to be updated.

python/lsst/dax/distribution/DataIngest.py
timj: In modern python you should use `subprocess.run()` directly. Much simpler API.

timj: Also note you'll get bytes back so will need to decode to str.

python/lsst/dax/distribution/DataGenClient.py
jgates108: The client sends back a list of _successfully_ completed chunks. The client needs to send back an empty list if no chunks were completed so the server knows what happened. I'll add a comment about ingestedChunks being destroyed in the process.

ctslater: Ok, I think I understand that part, but `clientReportChunksComplete()` says it "return a list of chunks that could not fit in the message." That's the behavior I think is confusing.

(When I said list of work above, I meant the "list of chunks to be reported as completed")

bin/edgefirsttst.py
jgates108: Removed None case.

bin/edgefirsttst.py
jgates108: Added to datgen.py configuration files (example_spec.py and fakeGenSpec.py).

python/lsst/dax/distribution/DataGenClient.py
jgates108: There's some discrepancy between PEP8 and the DM Python style guide when it comes to variable and function names in python. I prefer camel case, but I'll use whichever form people feel is more appropriate.

python/lsst/dax/data_generator/columns.py
```diff
@@ -213,3 +517,166 @@ def __call__(self, cell_id, length, prereq_row=None, prereq_tables=None):
             index_position += n_filter_visits
 
         return (objectId, ccdVisitId, psFlux, psFluxSigma)
+
+
+def tst_convertBlockToRows(log_msgs=True):
```
ctslater: All the tests should go into the `tests/` directory.

python/lsst/dax/data_generator/columns.py
```diff
@@ -1,24 +1,199 @@
...
+    Tuple of lists
+        Result of merging the two blocks.
+    """
+    newList = list()
+    # There's probably a more pythonic way to do this.
+    for j in range(len(block_a)):
+        newList.append(np.append(block_a[j], block_b[j]))
+    return tuple(newList)
+
+def equalBlocks(block_a, block_b):
```
ctslater: Should go with the tests in `tests/`.

python/lsst/dax/data_generator/columns.py
```diff
@@ -1,24 +1,199 @@
...
+    Return
+    ------
+    Tuple of lists
+        Result of merging the two blocks.
+    """
+    newList = list()
+    # There's probably a more pythonic way to do this.
+    for j in range(len(block_a)):
+        newList.append(np.append(block_a[j], block_b[j]))
+    return tuple(newList)
```
ctslater: I think this can be
```
return tuple(np.concatenate(x) for x in zip(block_a, block_b))
```

python/lsst/dax/data_generator/columns.py
```diff
@@ -28,179 +203,308 @@ def __call__(self, cell_id, length):
...
+        self.column_val = column_val  # arbitrary, but different from other columns
 
-    def __call__(self, cell_id, length, **kwargs):
-        mags = []
+    def __call__(self, chunk_id, length, seed, **kwargs):
+
+        np.random.seed(calcSeedFrom(chunk_id, seed, self.column_val))
+
+        # This needs to be made row by row not column by column, as
+        # row by row results in repeatable values when doing edges first.
```
ctslater, line 403: This is going to be really slow if it has to loop over each object. ForcedSource does this too, so MagnitudeGenerator on the Object table not dominate the runtime, but it's something to watch out for.

python/lsst/dax/distribution/DataGenClient.py
jgates108: I do want to differentiate between the client side of the connection and the server side of the connection and the server side (DataGenConnection may not be the best name for the class). The way the code is setup self._client.servSend() is obviously a mistake. I think that sort of thing has some value.  Maybe changing '_client' to '_cl_conn' ?

python/lsst/dax/distribution/DataGenClient.py
jgates108: The dictionary is part of the scheme to reliably send a random number of files across the connection without actually sending the number of files across the connection. The client starts asking for index 0 then index 1, and so on until there's no file for index x. This is explained in the server side of the code, I'll add a comment.

python/lsst/dax/distribution/DataGenClient.py
jgates108: I changed it to be a bit easier to read but kept the numerical index key
`for index, cfg_info in pCfgDict.items():`

python/lsst/dax/distribution/DataGenClient.py
jgates108: It's better to tell the server that nothing was created/ingested, otherwise the problems will register as a network failure instead of something wrong with this worker.

python/lsst/dax/distribution/DataGenClient.py
jgates108: Pretty much any error requires trying to abort the ingest transaction. Currently, if it isn't aborted here, the transaction needs to be tracked down and aborted by hand. Maybe raising the exception again after the _endTransaction call below?

python/lsst/dax/distribution/DataGenClient.py
jgates108: Sending the list of completed chunks back to the server is important every time through the loop, even if that list is empty

python/lsst/dax/distribution/DataGenClient.py
jgates108: I pulled that `while True:` loop out into a separate function. Re-sending a buffer until it is empty isn't that unusual in network programming.

python/lsst/dax/distribution/DataGenServer.py
jgates108: client_id is a string for flexibility and sending integers over a network connection in python is harder than sending text.

python/lsst/dax/distribution/DataGenServer.py
jgates108: Good catch

python/lsst/dax/distribution/DataIngest.py
jgates108: I'm finding trying to use subprocess.run really annoying. The cmd argument that works for subprocess.Popen doesn't work with subprocess.run. I've tried a few permutations and they've all failed, which I think is why I gave up on run in the first place.
`process = subprocess.run(cmd, cwd=cwd, shell=True, universal_newlines=True,
                                 stdout=subprocess.PIPE, stderr=subprocess.STDOUT)`
`process = subprocess.run([cmd, args], cwd=cwd, shell=True, universal_newlines=True,
                                 stdout=subprocess.PIPE, stderr=subprocess.STDOUT)`

I'm trying to execute commands like 
`python ~/work/dax_data_generator/bin/datagen.py  --chunk 1201 --visits 30 --objects 1000 gencfg.py`
    subprocess.run consistently complains that it can't find datagen.py
Or 
` sph-partition -c /home/jgates/work/dax_data_generator/python/lsst/dax/distribution/fakeData/partitionCfgs/Object.cfg --mr.num-workers 1 --out.dir outdirObject --in chunk0_CT_Object.csv --in chunk402_CT_Object.csv --in chunk401_CT_Object.csv --in chunk400_CT_Object.csv --in chunk404_EO_Object.csv --in chunk403_CT_Object.csv`
Reconstructing them as lists is a nuisance.

Having the whole command line in one string makes it really easy to debug failures as I just cut and paste the string in a terminal.

python/lsst/dax/distribution/DataGenClient.py
jgates108: It probably is worth a try. Doing it this way made testing it fairly easy. There is also the idea of running multiple datagen.py at once, which I still haven't completely ruled out. Having it run separately allows for real parallelism, which is can be tricky within a single python process.

python/lsst/dax/distribution/DataGenServer.py
jgates108: I like None in that it is obvious that client_addr was never defined instead of it being set by an empty string in log file. If this is a python faux pas, then changing it isn't a big deal. I will add a comment that it should be string.

For both of these (client_id = '-1' and client_addr = None), the value serves as an indicator that it has not been defined.  Consistency is nice, so which is better in python land, '-1' or None?

python/lsst/dax/distribution/DataGenConnection.py
jgates108: I added lines between the functions for this file.

python/lsst/dax/distribution/DataGenClient.py
ctslater: Yeah, my issue is that python is not very good at managing subprocesses, and trying to pull off parallelism with subprocesses is a real pain (I've tried, and regretted it). The `multiprocessing` module is probably the best way to get multiple cores running on the data generator side. It's a bit limited, but I'd argue not any more limited than subprocesses. If that and/or multiple DataGenClients is not sufficient parallelism then I would consider some sort of external manager program.

python/lsst/dax/distribution/DataIngest.py
timj: You may well find that `~` is not expanded by the subprocess command depending on the shell setting. It should be able to take a single string because it is documented to take the normal Popen commands but with a much simplified interface.

Having all the commands separately is much more secure and automatically deals with any problems with shell quoting.

python/lsst/dax/distribution/DataGenConnection.py
```diff
@@ -0,0 +1,730 @@
...
+        diff = self.compareChunkLists(allChunks, sentChunks)
+        if len(diff) != 0:
+            print("large range chunkMsg failed", diff)
+            success = False
+        if success is None: success = True
+        print("testMethods success=", success)
+        return success
+
+#####################################################
+### Following only for testing
```
ctslater: Tests to `tests/`.

python/lsst/dax/distribution/DataIngest.py
```diff
@@ -0,0 +1,340 @@
...
+        print("ERROR failed to send database configuration to ingest.")
+    # Ingest the test Object tables schema. This will fail if the
+    # database already exists.
+    if not ingest.sendTableSchema("fakeIngestCfgsTest/test102_Object.json"):
+        print("ERROR failed to send Object table schema")
+    # Start a transaction
+    i_transaction = IngestTransaction(ingest, 'test102')
+    transaction_status = None
+    try:
+        with i_transaction as t_id:
```
ctslater: It's a little nonstandard not to do `with IngestTransaction(...) as blah:`. I _think_ if you want to be able to set `abort` inside the transaction object, you can have `IngestTransaction.__enter__()` return `self`?

python/lsst/dax/distribution/README.txt
```diff
@@ -0,0 +1,71 @@
+
```
ctslater: I would merge this into the top level README.

python/lsst/dax/distribution/fakeGenSpec.py
```diff
@@ -0,0 +1,30 @@
+import lsst.dax.data_generator.columns as columns
```
ctslater: This is really a config file, so putting it inside `python/` is confusing (it could potentially get imported by python, and it's not meant for that). I'd put it with other config files.

python/lsst/dax/distribution/fakeIngestCfgs/fakedb.json
```diff
@@ -0,0 +1,7 @@
+{"database":"fakedb",
```
ctslater: I'd put all these configs somewhere outside of `python/`. Also I don't understand the meaning of "fake" in `fakeIngestCfgs`.

bin/datagen.py
```diff
@@ -2,20 +2,32 @@
...
+#   python bin/datagen.py  --chunk 3525 --visits 30 --objects 1000 example_spec.py
+# edge first only the edge:
+#   python bin/datagen.py --edgeonly --chunk 3525 --visits 30 --objects 1000 example_spec.py
 if __name__ == "__main__":
 
     parser = argparse.ArgumentParser()
     parser.add_argument("--chunk", type=int, required=True)
     parser.add_argument("--objects", type=int, required=True)
     parser.add_argument("--visits", type=int, required=True)
+    parser.add_argument("--edgeonly", action="count", default=0)
```
ctslater: `parser.add_argument("--edgeonly", action="store_true")`.

python/lsst/dax/distribution/DataGenServer.py
```diff
@@ -0,0 +1,513 @@
...
+            success, status, r_json = self._ingest.publishDatabase(self._db_name)
+            if success:
+                print("Published", self._db_name)
+            else:
+                print("ERROR failed to publish", self._db_name, status, r_json)
+        else:
+            print("Not publishing due to incomplete data/creation/ingestion")
+
+
+def testA():
```
ctslater: This should go in a script in `bin/`

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,799 @@
...
+                # chunks to create would just cause more problems.
+                if len(ingestedChunks) == 0:
+                    print("ERROR no chunks were successfully ingested, ending program")
+                    loop = False
+
+                # Client sends the list of completed chunks back
+                self._sendIngestedChunksToServer(ingestedChunks)
+
+
+def testC():
```
ctslater: This should go in `bin/`.

python/lsst/dax/distribution/DataGenConnection.py
```diff
@@ -0,0 +1,730 @@
...
+                completed_chunks.extend(completedC)
+            # compare with original chunk list
+            diff = serv.compareChunkLists(completed_chunks, self.chunkListA)
+            print("final compare diff=", diff)
+            if len(diff) != 0:
+                self.success = False
+                raise RuntimeError("mismatch in sent vs received lists", self.name, diff)
+            self.warnings += serv.warnings
+        print("ServerTestThrd.run finished")
+        if self.success == None: self.success = True
```
kennywlo: Should use is operator for comparison with None.

python/lsst/dax/distribution/DataGenConnection.py
```diff
@@ -0,0 +1,730 @@
...
+                print("chunks read success")
+            self.warnings += client.warnings
+            # Send the list of completed chunks back
+            completedChunks = chunkListARecv.copy()
+            while len(completedChunks) > 0:
+                completedChunks = client.clientReportChunksComplete(completedChunks)
+
+
+        print("ClientTestThrd.run finished")
+        if self.success == None: self.success = True
```
kennywlo: Should use is operator for comparison with None.

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,799 @@
...
+            # incrementing the index and asking for a file until the server
+            # responds with an empty file name.
+            pCfgIndex = 0
+            pCfgDict = {}
+            pCfgName = "nothing"
+            while not pCfgName == "":
+                self._cl_conn.clientReqPartitionCfgFile(pCfgIndex)
+                indx, pCfgName, pCfgContents = self._cl_conn.clientRespPartionCfgFile()
+                if indx != pCfgIndex:
+                    self.success = False
```
kennywlo: Instance variable `success` should be defined in __init__.

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,799 @@
...
+
+    def run(self):
+        """Connect to the server and do everything until the server
+        runs out of chunks for this client to generate and ingest.
+        """
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.connect((self._host, self._port))
+            self._cl_conn = DataGenConnection(s)
+            self._cl_conn.clientReqInit()
+            self._name, self._gen_arg_str, self._cfg_file_contents, ingest_dict = self._cl_conn.clientRespInit()
```
kennywlo, line 695: Instance variable `_name` should be defined in __init__.

python/lsst/dax/data_generator/columns.py
```diff
@@ -28,179 +203,308 @@ def __call__(self, cell_id, length):
...
+        ra_out = []
+        for ra in ra_centers:
+            while ra < 0.0:
+                ra += 360.0
+            while ra >= 360.0:
+                ra -= 360.0
+            ra_out.append(ra)
+        return (ra_out, dec_centers)
+
+    def __call__(self, chunk_id, length, seed, edge_width, edge_only, prereq_tables=None, **kwargs):
```
kennywlo: Signature does not match that of base method in 'ColumnGenerator'.

python/lsst/dax/distribution/DataGenConnection.py
```diff
@@ -0,0 +1,730 @@
...
+        if len(diff) == 2:
+            if x not in diff or val not in diff:
+                print("compare b has more elements val failed", diff)
+                success = False
+        else:
+            print("compare b has more elements len failed", diff)
+            success = False
+        # Test a reasonable length range for ChunkMsg methods
+        b = range(1,110)
+        testMsg, usedChunks = self._buildChunksMsg(b)
```
kennywlo, line 512: Expected type 'List(int)', got 'range' instead.

ctslater (commented):
I mostly have file organization comments, plus some responses that got caught in github's "review" and it wouldn't let me send separately.

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -0,0 +1,799 @@
...
+                             edge_only=False, use_targ_path=True))
+            for fn in fList:
+                if os.path.exists(fn):
+                    print("removing file", fn)
+                    if not self.removeFile(fn):
+                        print("ERROR remove failed", fn)
+                        return False
+        return True
+
+    def _fillChunkDir(self, chunk_id, neighborChunks):
```
kennywlo: variable name should be lowercase.

python/lsst/dax/data_generator/generator.py
```diff
@@ -75,16 +108,23 @@ def _add_to_list(self, generated_data, output_columns, split_column_names):
                 assert ValueError, ("Column name implies multiple returns, "
                                     "but generator only returned one")
 
-    def make_chunk(self, chunk_id, num_rows=None):
+    def make_chunk(self, chunk_id, num_rows=None, seed=1, edge_width=0.017, edge_only=False):
```
kennywlo, line 111: Above: redundant ().  Should assert be a raise instead?

kennywlo (commented):
Outside functionality, I can see that the code needs to perform well, given its purpose.  For the multiple nested loops, I would suggest putting in tracing for either # of items being processed, or timestamps to help track the execution complexity on a regular basis.  This  info would also help determine which code sections can benefit from multi-processing within the Python context.   If not done already, a run with code coverage turned on will help in this regard.

python/lsst/dax/distribution/DataGenClient.py
jgates108: I agree with you, but I think I'd like to make a different ticket to call the class directly. I have to change the communication protocol and there are some other things I'd like to change in DataGenerator (mostly having to do with what's put on the command line and what is in example_spec.py ) before altering the protocol.

python/lsst/dax/distribution/DataIngest.py
jgates108: I did the same thing I did yesterday and today it worked. I'm at a loss.

python/lsst/dax/distribution/DataGenClient.py
ctslater: Putting that on a separate ticket sounds good to me.

python/lsst/dax/data_generator/columns.py
jgates108: That seems more pythonic.

python/lsst/dax/data_generator/columns.py
jgates108: I've got 'tst_RaDecGenerator(log_msgs=True, every_nth=75)' in columns.py. I have it called once from 'tests/', which tests a fairly sparse set of chunks and one in 'bin/edgetest.py' that checks far more chunks and is slow. That's why I left the definition here. I occasionally want to run the slow version as it's more likely to turn up possible issues with random number generation. Is there a reasonable way to still access this from 'tests/'?

python/lsst/dax/data_generator/columns.py
jgates108: Generating it by column requires knowing how many rows are in the edge only version, and there's no mechanism for that at this point. With luck this will be fast enough.

python/lsst/dax/distribution/fakeGenSpec.py
jgates108: Where are configuration files usually put?

python/lsst/dax/distribution/fakeIngestCfgs/fakedb.json
jgates108: 'fake' just means the data is made up, and all of the configuration files are for 'fakedb'. All the configuration files are pretty tightly linked. If something changes in one, there probably need to be changes to most of the configuration files. The other systems want similar information in different formats. If there's a different database to be created, it will need an entire new set of config files and directories. 

I don't really expect to need that different databases, but I expect there will be a kpm50db that will not be suitable for quick testing on a lone pc, and that will have 'kpm50IngestCfgs'.

ctslater: I think if there was a `config/` directory in the repository root, then `config/fakedb/` would be an understandable path. Right now I keep reading it as ingest itself being "fake", somehow. (I also wonder if there's something more specific to say than just "fakedb", since the other databases this generates will also be "fake")

There's not really a python standard for configuration data (as far as I know), so anywhere that's distinct from the python code should be fine.

python/lsst/dax/data_generator/columns.py
ctslater: Breaking out the test code into a separate file like `python/lsst/dax/data_generator/testing.py` would seem reasonable to me. That would also confine them to their own module (`lsst.dax.data_generator.testing`).

python/lsst/dax/data_generator/generator.py
jgates108: I agree with this but more in the terms of time spent in various stages of data generation and ingestion. I think it's much more useful if I have the data produced and sent back to the server, which also ties into calling DataGenerator directly and the protocol changes. I created DM-26282 to cover both of these.

python/lsst/dax/distribution/DataGenClient.py
jgates108: DM-26282

python/lsst/dax/distribution/DataIngest.py
jgates108: I wound up not using this code outside of testing. If there's reliability problems with clients, it will be worth having the server close the individual transactions so that the server really knows which files have been ingested. I'm hoping client reliability wont be that much of an issue.

python/lsst/dax/distribution/fakeIngestCfgs/fakedb.json
jgates108: The fake database is going to be visible along side wise and stripe 82 in qserv and it's not really deserving of of a name like synthetic as it overwhelmingly consists of uncorrelated random numbers. It's only good for bulk testing of qserv and I wouldn't want anyone to think it's something interesting, hence 'fake'.

example_spec.py
```diff
@@ -4,6 +4,7 @@
 num_stripes = 200
 num_substripes = 5
 chunker = Chunker(0, num_stripes, num_substripes)
+edge_width = 0.017  # degrees, must be >= overlap
```
iagaponenko, line 7: (assuming this is for "overlap") Why not `0.01667` like in other examples of the project?

python/lsst/dax/data_generator/generator.py
```diff
@@ -64,27 +98,33 @@ def _add_to_list(self, generated_data, output_columns, split_column_names):
...
         else:
             output_columns[split_column_names[0]].append(generated_data)
             if(len(split_column_names) > 1):
-                assert ValueError, ("Column name implies multiple returns, "
-                                    "but generator only returned one")
+                raise ValueError("Column name implies multiple returns, "
+                                 "but generator only returned one")
 
-    def make_chunk(self, chunk_id, num_rows=None):
+    def make_chunk(self, chunk_id, num_rows=None, seed=1, edge_width=0.017, edge_only=False):
```
iagaponenko, line 111: Is `0.017` for "overlap"? If so then why not 0.01667 like in other examples of the project?

python/lsst/dax/distribution/DataIngest.py
```diff
@@ -0,0 +1,297 @@
...
+    """This class is used to communicate with the ingest system using
+    json formated requests and responses.
+
+    Parameters
+    ----------
+    host : str
+        Data ingest server host.
+    port : int
+        Data ingerst server port number.
+    user, auth_key : str, optional
```
iagaponenko: Parameter "user" is not needed, and it's confusing. No users exist in the Ingest system. Please, remove it!

python/lsst/dax/distribution/DataIngest.py
```diff
@@ -0,0 +1,297 @@
...
+        User name and authorization key.
+    """
+
+    def __init__(self, host, port, user='', auth_key=''):
+        self._host = host
+        self._port = port
+        self._user = user
```
iagaponenko: Please, remove `user`.

python/lsst/dax/distribution/DataIngest.py
```diff
@@ -0,0 +1,297 @@
...
+        out = ("host=" + self._host + ":" + str(self._port)
+             + " user=" + self._user + "auth_key:****")
+        return out
+
+    def isIngestAlive(self):
+        """Test if ingest system is alive.
+        """
+        url = self._base_url + 'meta/version'
+        print("ingest url=", url)
+        response = requests.get(url)
```
iagaponenko: In all requests to REST services you should first check for the HTTP status code before drilling down onto the `JSON` response. Here is how this is done:
```
if response.status_code != 200: print "error: request failed with HTTP code: " + response.status_code 
```
Note, this applies to *ALL* requests, not just this one. Otherwise, you will get an exception at the next line in case of a error.

python/lsst/dax/distribution/DataIngest.py
```diff
@@ -0,0 +1,297 @@
...
+        r_json = response.json()
+        if not r_json['success']:
+            print('error: ' + r_json['error'])
+            return False
+        print('ingest version=', r_json['version'])
+        return True
+
+    def _postToIngest(self, ingest_cmd, data_json):
```
iagaponenko: Code duplication in _postToIngest` and `_putToIngest` as you may add the HTTP method name 'POST', 'PUT', 'GET' or 'DELETE' as  parameter to the general purpose method:
```
def _requestToIngest(self, method, service, data_json):
   ...
```

python/lsst/dax/distribution/DataIngest.py
```diff
@@ -0,0 +1,297 @@
...
+            success = False
+            return success, status_code, None
+        r_json = response.json()
+        if not r_json['success']:
+            print('ERROR put url=', url, 'status=', status_code, 'data=', data_json,
+                  'r_json=', r_json)
+            success = False
+        return success, status_code, r_json
+
+    def sendDatabase(self, db_file_path):
```
iagaponenko: Yes, indeed, this method is sending some JSON object to the corresponding service. However, this is not this operation is meant to do. The right name for the method should be `registerDatabase` or `createDatabase`.

python/lsst/dax/distribution/DataIngest.py
```diff
@@ -0,0 +1,297 @@
...
+        with open(db_file_path, 'r') as db_f:
+            data = db_f.read()
+            data_json = json.loads(data)
+        success, r_json = self._postToIngest('ingest/database', data_json)
+        if not success:
+            print('ERROR while sending databaae ', db_file_path, "r_json=", r_json)
+            return False
+        return True
+
+    def sendTableSchema(self, schema_file_path):
```
iagaponenko: See my comment for method `sendDatabase`. This method should be named as `registerTable` or `createTable`.

python/lsst/dax/distribution/DataIngest.py
```diff
@@ -0,0 +1,297 @@
...
+        Return
+        ------
+        success : bool
+            True if the operation was successful.
+        status : int
+            Status code from the put operation.
+        r_json : json or None
+            json data from the put operation if status was 200.
+        """
+        cmd = 'ingest/trans/' + str(transaction_id) + '?abort='
```
iagaponenko: You may rewrite this as a single line in Python 2:
```
cmd = 'ingest/trans/%d?abort=%d' % (transaction_id,int(abort),)
```
Or, like this in Python 3:
```
cmd = 'ingest/trans/{}?abort={}'.format(transaction_id,int(abort),)
```

python/lsst/dax/distribution/DataIngest.py
```diff
@@ -0,0 +1,297 @@
...
+        host : str
+            Host name of the ingest worker.
+        port : int
+            Port number of the ingest worker.
+        """
+        cmd = 'ingest/chunk'
+        jdata = {"transaction_id":transaction_id,"chunk":chunk_id,"auth_key":self._auth_key}
+        success, r_json = self._postToIngest(cmd, jdata)
+        if not success:
+            print("ERROR failed to get chunk target address", jdata, " r_json=", r_json)
```
iagaponenko, line 198: This printout duplicates what the exception does. And it bloats your code here. I presume you may just print the message encoded into the exception at a point where you will be catching it upstream.

python/lsst/dax/distribution/DataIngest.py
```diff
@@ -0,0 +1,297 @@
...
+            The name of the table that f_path data should be added to.
+        f_path : str
+            Full path to the file to ingest.
+
+        Note
+        ----
+        This calls the external program 'qserv-replica-file-ingest' to
+        actually send the file.
+        """
+        cmd = ('qserv-replica-file-ingest FILE ' + host + ' ' + str(port) + ' '
```
iagaponenko, line 226: For the record, this operation may be less efficient than sending many files in batch mode as explained here: https://confluence.lsstcorp.org/pages/viewpage.action?pageId=133333850#UserguidefortheQservIngestsystem(APIversion1)-Uploadingchunkcontributions and here: https://confluence.lsstcorp.org/display/DM/Ingest%3A+11.2.2.+qserv-replica-file-ingest
Though, this may require to improve your workflow

python/lsst/dax/distribution/DataIngest.py
```diff
@@ -0,0 +1,297 @@
...
+            + str(transaction_id) + ' ' + table + ' P ' + f_path + ' --verbose --columns-separator=TAB')
+        print("cmd=", cmd)
+        process = subprocess.run(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+        out_str = str(process.stdout)
+        if process.returncode != 0:
+            print("ERROR sendChunkToTarget cmd=", cmd, "out=", out_str)
+            raise RuntimeError("ERROR sendChunkToTarget cmd=" + cmd + " out=" + str(out_str))
+        return process.returncode, out_str
```
iagaponenko, line 234: It makes no sense to return the completion code from this method as it would always be 0. Otherwise you will throw exception `RuntimeError`.

example_spec.py
jgates108: It just leaves a small buffer (0.00033 degrees) for creating overlaps.

python/lsst/dax/distribution/DataIngest.py
jgates108: At some point, aren't we going to need to ingest datasets owned by individual observers other than LSST? How are the data owner and access rights going to be passed in?

python/lsst/dax/distribution/DataIngest.py
jgates108: Both of the following return "ingest/trans/543?abort=True" which doesn't match a 0 or 1 value for abort.
cmd = 'ingest/trans/{}?abort={}'.format(transaction_id,abort)
cmd = f'ingest/trans/{transaction_id}?abort={abort}'

This does work
cmd = 'ingest/trans/%d?abort=%d' % (transaction_id,abort)

python/lsst/dax/distribution/DataIngest.py
iagaponenko: We don't have any plans for that yet. So, I think it's premature to plan for that. Besides, at some point there will be an "official" Python module for interacting with the REST API.

python/lsst/dax/distribution/DataIngest.py
iagaponenko: My bad. I have corrected my original suggestion into:
```
cmd = 'ingest/trans/%d?abort=%d' % (transaction_id,int(abort),)
cmd = 'ingest/trans/{}?abort={}'.format(transaction_id,int(abort),)
```
This should work.

bin/datagen.py
ctslater: Since `args.edgeonly` is now a boolean, you don't need the separate `edge_only` variable down below.

python/lsst/dax/distribution/DataIngest.py
jgates108: I did it this because it made the rest of the code much simpler and it will take a substantial code change to alter it. I guess a reasonable question is how much more efficient? I expect files to be many megabytes or gigabytes in most cases and I hope the overhead to ingest a file is very small in comparison. The need for this is getting to be urgent, so I'd like to see if optimization is needed before spending time on it.

iagaponenko: It all depends on the total number of files and file sizes, and also how you're going to submit your contributions. If you do everything from inside a container then an overhead for launching the `qsev-replica-file-ingest` binary may be small. However if you're going to launch a Docker container for a single file ingest then it would be a very different story. 
Your choices (which of those two options to use: `FILE` or `FILE-LIST`) would also depend on an organization of your workflow. From my reading of your comments in the JIRA ticket on tis code, it looks like your workflow is around each chunk:
1) you have a list of chunks to be generated
2) you have multiple "clients" each asking your server the next available chunk to process
3) and each clients is working on one chunk at a time, generates the "fake" data, partitions the chunk file to make overlaps and ingests the resulting products into Qserv

In this model `FILE` looks like a natural choice.  The main benefit of your approach is that it won't require any significant temporary storage to accumulate a lot of data during steps 1 and 2 above.
An alternative approach would be to let each worker generate N chunks, partition them into some output folder and then initiate the ingest phase using `FILE-LIST`. This approach will require you to be cautious regarding the amount of free space available in your temporary storage area.
In the end it's your call. You will see how it all works when you will be doing large-scale generation/ingest. You may always optimize your workflow later.
Just one last word of caution: please, avoid starting too many "super-transactions" in your workflow. It would be reasonable to keep the total number of transactions you have during the whole ingest campaign under 100 or so. See some notes on this subject at: https://confluence.lsstcorp.org/pages/viewpage.action?pageId=133333850#UserguidefortheQservIngestsystem(APIversion1)-Multipletransactions

iagaponenko:
Looks good to me

PR #2, opened 2020-08-25
Tickets/dm 26409

jgates108:


container/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+FROM qserv/qserv:deps_20200729_0234
+
+USER 1000
+
+RUN source /qserv/stack/loadLSST.bash && \
+    setup -t qserv-dev qserv_distrib && \
+    setup partition && \
+    git clone https://github.com/lsst/sphgeom.git && \
+    cd sphgeom && setup -k -r . && scons -j10 install && \
+    cd .. && \
```
iagaponenko: I would suggest this instead:
```
cd -
```
This would set `PWD` to a previous folder regardless of a depth:
```
cd abc/d/e/f
cd -
```

container/Dockerfile
```diff
@@ -0,0 +1,15 @@
+# Using the Qserv development toolchain
+FROM qserv/qserv:deps_20200729_0234
+
+USER 1000
+
+RUN source /qserv/stack/loadLSST.bash && \
+    setup -t qserv-dev qserv_distrib && \
+    setup partition && \
```
iagaponenko: You don't need this as `partition` will be added by `setup -t qserv-dev qserv_distrib`

container/Dockerfile.tmpl
```diff
@@ -0,0 +1,11 @@
+# Using the Qserv development toolchain
```
iagaponenko, line 1: I have noticed that `Dockerfile` and `Dockerfile.tmpl` are substantially different. I would argue that this repos should only had `Dockerfile.tmpl` of which you would produce the former for a given branch. In this case `Dockerfile` had to be merged into `Dockerfile.tmpl` and be removed from the repo.

scripts/README.md
```diff
@@ -0,0 +1,19 @@
...
+```
+docker pull qserv/dax_data_generator:tools
+```
+
+To start the client (replace 'tools' with 'tools-DM-<branch#>'):
+```
+docker run --network=host --rm -it -u 1000:1000 qserv/dax_data_generator:tools \
+  /bin/bash -c "/home/qserv/dax_data_generator/scripts/run_datagenclient.sh"
+```
+
```
iagaponenko: Here I would also add instructions for passing an external configuration to the client and server using Docker's option `-v`:
```
docker run ... -v ./dax_data_generator/config/fakedb:/home/qserv/dax_data_generator/config/fakedb:ro ...
```
This would override the default configuration provided by the git package.
Note that one can also do the per-file change like:
```
-v fakeGenSpec.py:/home/qserv/dax_data_generator/config/fakedb/fakeGenSpec.py:ro
-v serverCfg.yml:/home/qserv/dax_data_generator/config/fakedb/serverCfg.yml:ro
```
etc.

scripts/run_datagenclient.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+source /qserv/stack/loadLSST.bash
+setup -t qserv-dev qserv_distrib
+cd /home/qserv/sphgeom
+setup -k -r . -t qserv-dev
+cd /home/qserv/dax_data_generator/
+python bin/datagenclient.py "$@"
```
iagaponenko, line 7: Don't you need to add:
```
setup -k -r .
```
before launching the client?

scripts/run_datagenserver.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+source /qserv/stack/loadLSST.bash
+setup -t qserv-dev qserv_distrib
+cd /home/qserv/dax_data_generator/
+python bin/datagenserver.py "$@"
```
iagaponenko, line 5: Don't you need to add:
```
setup -k -r .
```
before launching the server?

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -63,7 +63,8 @@ def __init__(self, host, port, target_dir='fakeData', chunks_per_req=5):
         self._cl_conn = None # DataGenConnection
         self._cfg_file_name = 'gencfg.py' # name of the local config file for the generator
         self._cfg_file_contents = None # contents of the config file.
-        self._datagenpy = '~/work/dax_data_generator/bin/datagen.py' # TODO: MUST stop hard coding this
+        #self._datagenpy = '~/work/dax_data_generator/bin/datagen.py' # TODO: MUST stop hard coding this &&&
+        self._datagenpy = '~/dax_data_generator/bin/datagen.py' # TODO: MUST stop hard coding this &&&
```
iagaponenko, line 67: This is very unsafe. There are two reasons for this First of all, `~` implies the current user which may potentially be different from a location where you put the package. The second problem is that your package may be installed at some other location.
Assuming that the current script is imported into one of your binary tools deployed at the package's `bin/` folder you may try to figure out an absolute path of that folder by:
```
import os.path
import sys
bindir = os.path.dirname(os.path.abspath(sys.argv[0]))
if bindir == '': basedir = '.'
```
After that you could do:
```
self._datagenpy = '{}/datagen.py'.format(bindir)
```

container/README.md
```diff
@@ -0,0 +1,30 @@
...
+repo will be IGNORED by the Dockerfile.
+
+Assuming the container is build off the `master` branch of the package, do tis
+from the top directory of the package:
+```
+BRANCH=master
+cat container/Dockerfile.tmpl | sed 's/{{BRANCH}}/'${BRANCH}'/g' > container/Dockerfile
+docker build -t qserv/dax_data_generator:tools -f container/Dockerfile .
+docker push qserv/dax_data_generator:tools
+```
```
iagaponenko, line 27: I would suggest writing a shell script `container/build.sh` which would incorporate 4 lines above into something like this:
```
#!/bin/bash

set -e

CONTAINER_TOOLS_DIR="$0"
USAGE="
Usage:
    ${CONTAINER_TOOLS_DIR} <branch>

Where:
  <branch> - the name of the Git branch to build the container from. Note that if you
             have a local copy of the branch then commit and push your changes to
             GitHub before building the container.
             The branch name will also be used to create a custom tag for the
             container. For example, branch 'tickets/DM-26409' will map to
             tag 'qserv/dax_data_generator:tickets-DM-26409'. And branch 'master' will
             make tag 'qserv/dax_data_generator:master'.
"
BRANCH="$1"
if [ -z "${BRANCH}" ]; then
    >&2 echo $USAGE
    exit 1
fi
TAG="qserv/dax_data_generator:$(echo $BRANCH | tr '/' '-')"
cd $CONTAINER_TOOLS_DIR
cat Dockerfile.tmpl | sed 's/{{BRANCH}}/'${BRANCH}'/g' > Dockerfile
docker build -t $TAG -f Dockerfile .
docker push $TAG
rm Dockerfile
```

iagaponenko:
I think the code still needs a few improvements. I have made my suggestions above.

scripts/README.md
jgates108: datagenclient.py only gets host and port of the server from command line arguments. All of the other configuration values are provided by datagenserver.py and sent over the network connection.

The server doesn't really need to run in a container, but it can be. Since there really should be only one datagenserver in the system it's easier to set it up outside of a container. There are a couple of server arguments that are most easily handled on the command line, like don't try to register the schema with ingest since it has already been done is '-s' or '-k' to skip ingest entirely. What you propose makes sense for the server in a container, but we don't need that at this point.

scripts/run_datagenclient.sh
jgates108: There's no eups directory in dax_data_generator. Maybe there should be?

python/lsst/dax/distribution/DataGenClient.py
jgates108: The other ticket I'm working on is largely about removing this bit code and calling DataGenerator in python without calling an outside procedure. That is done and this hard coded line is gone in the other ticket.

scripts/README.md
iagaponenko: I missed this fact (that clients don't require any external configuration). However, you the proposed  Docker's volume mounting option will be required to run the server.

I doubt your Python code would run outside of the Docker environment since it requires LSST Stack environment and Python 3. None of the target machines (Qserv clusters at NCSA) where the code is meant to be run has such environment.

scripts/run_datagenclient.sh
iagaponenko: Eventually, a correct version of `sphgeom` will be provided by the Stack. If your code works well after you do `setup` like you do it now then you will not need to have an explicit dependency onto  `sphgeom` in the future.

python/lsst/dax/distribution/DataGenClient.py
iagaponenko: 👍

iagaponenko (approved):

container/README.md
jgates108: I added a build.sh file, it's a bit different than this one.

scripts/README.md
jgates108: I'm adding the instruction. The config files are complex, so it may be easier to setup the stack and everything on one computer to get the configuration files correct. I suspect I'll probably try to get configurations working locally for a dozen chunks before trying to set them up on a cluster.

scripts/run_datagenclient.sh
fritzm: Yes -- given the recent merges to sphgeom, is it necessary to explicitly `setup` the version in /home/qserv/sphgeom, or can you just use the version you get from setting up `qserv_distrib` above?

fritzm (approved):

scripts/run_datagenclient.sh
jgates108: the stack doesn't yet have the correct sphgeom. We'll need to come back to this when it does.

PR #3, opened 2020-08-31
Tickets/dm 26282

jgates108:


python/lsst/dax/data_generator/timingdict.py
```diff
@@ -0,0 +1,146 @@
...
+
+    Notes
+    -----
+    While this is intended to store timing information, the only hard
+    requirement is that all the keys are strings and all the values are
+    floats.
+    The key "count" is reserved for the total count.
+    """
+
+    def __init__(self, timing_dict=None):
```
ctslater: A few inter-related issues here: it doesn't look like you actually ever use this timing_dict parameter outside of the tests, so I would remove that parameter. Copy constructors aren't really a common python pattern, and I think it makes it harder to read the code when there's a `timing_dict` and a `TimingDict` both floating around.

python/lsst/dax/data_generator/timingdict.py
```diff
@@ -0,0 +1,146 @@
...
+            Starting time for a measurent.
+        """
+        end_time = time.time()
+        val = end_time - start_time
+        self.add(key, val)
+
+    def increment(self):
+        """Increment "count"
+        """
+        self.add("count", 1.0)
```
ctslater: Why not just make this an instance variable instead of having to stick it in the dictionary? Also, an int.

python/lsst/dax/data_generator/timingdict.py
```diff
@@ -0,0 +1,146 @@
...
+        the 'other' value.
+        """
+        for key, val in other.timing_dict.items():
+            if key in self.timing_dict:
+                self.timing_dict[key] += val
+            else:
+                self.timing_dict[key] = val
+
+    def report(self):
+        """Get a somewhat well formed string of the contents of this object.
```
ctslater, line 117: I'm pretty confused by this function overall; it seems like overloading the dictionary to also include a count variable makes every access to the dictionary much more complicated, and creates some confusing patterns.

python/lsst/dax/data_generator/timingdict.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        st = "Times\n"
+        COUNT = 'count' # reserved string
+        count = None
```
ctslater: I'm pretty opposed to having capitalization-differentiated variables, but I think the real `count` variable should be an instance variable anyways and this would all be way simpler.

python/lsst/dax/data_generator/timingdict.py
```diff
@@ -0,0 +1,146 @@
...
+        width = 5
+        if COUNT in self.timing_dict:
+            if self.timing_dict[COUNT] != 0:
+                count = self.timing_dict[COUNT]
+        sum = 0.0
+        for key, val in self.timing_dict.items():
+            if key == COUNT:
+                continue
+            if (len(key) > width):
+                width = len(key)
```
ctslater: `width = max(len(x) for x in self.timing_dict.keys())`, no loop necessary.

python/lsst/dax/data_generator/timingdict.py
```diff
@@ -0,0 +1,146 @@
...
+            sum += val
```
ctslater, line 123: `sum` is a builtin, bad to shadow it.
`timing_sum = sum(self.timing_dict.values())` if you take out the COUNT element.

python/lsst/dax/data_generator/timingdict.py
```diff
@@ -0,0 +1,146 @@
...
+
+        Parameters
+        ----------
+        key : str
+            The key name.
+        val : float
+            Value to add to the existing value. Existing value is 0 for
+            undefined keys.
+        """
+        if key in self.timing_dict:
```
ctslater: Making this a `collections.defaultdict` with a default of zero would avoid having to special-case the initial value.

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -431,15 +463,14 @@ def _generateChunk(self, chunk_id, edge_only=False):
...
-            options += " --edgeonly "
-        cmdStr = ("python " + self._datagenpy + options +
-            " --chunk " + str(chunk_id) + " " + self._gen_arg_str + " " + self._cfg_file_name)
-        genResult, genOut = self.runProcess(cmdStr)
-        if genResult != 0:
-            print("ERROR Generator failed for", chunk_id, " cmd=", cmdStr, "out=", genOut)
+        # Genrate the chunk csv files.
+        try:
+            self._datGenChunk(chunk_id, edge_only)
+        except IndexError as ie:
```
ctslater, line 470: Where does this `IndexError` get thrown? It might be easier to catch the `IndexError` near to where it occurs and re-throw a more general `RuntimeError`, so the higher levels only need to look for one thing.

python/lsst/dax/distribution/DataGenConnection.py
```diff
@@ -418,6 +426,64 @@ def clientRecvChunks(self):
...
+        print("servRecvTiming C_TIMDCT")
+        msg_id, msg, msg_len = self._recv_msg()
+        if msg_id != self.C_TIMDCT:
+            self.warnings += 1
+            raise DataGenError(f'ERROR servRecvTiming {str(msg_id)}:{str(msg_len)} {msg}')
+        msg_split = msg.split(self.COMPLEXSEP)
+        time_dict = {}
+        a_key = True
+        last_key = ""
+        # msg_split contains an alternating sequence of key then value
```
ctslater: Can you use distinct separators between (key, value) pairs vs. between the key and the value? The for-loop flip-flop would be nice to avoid.

python/lsst/dax/distribution/DataGenConnection.py
```diff
@@ -418,6 +426,64 @@ def clientRecvChunks(self):
...
+        count = 0
+        for tkey, tval in timing_dict.items():
+            count += 1
+            if count > 50:
+                break
+            sep = c_sep
+            if first:
+                first = False
+                sep = ''
+            time_msg += sep + str(tkey) + c_sep + str(tval)
```
ctslater: The different separators going on is hard to keep track of, but I think the basic functionality you're going for is something like `";".join(f"{k},{v}" for (k,v) in timing_dict.items()[:50])`

python/lsst/dax/distribution/DataGenConnection.py
```diff
@@ -418,6 +426,64 @@ def clientRecvChunks(self):
...
+        # msg_split contains an alternating sequence of key then value
+        for m in msg_split:
+            if a_key:
+                a_key = False
+                last_key = m
+            else:
+                a_key = True
+                val = float(m)
+                time_dict[last_key] = val
+        return time_dict
```
ctslater: It's pretty confusing for some parts of the code to have a `timing_dict` that's just a dict and some parts `timing_dict` is a `TimingDict`.

python/lsst/dax/data_generator/timingdict.py
```diff
@@ -0,0 +1,146 @@
...
+        if self.timing_dict is None:
+            self.reset()
+
+    def __repr__(self):
+        return str(self.timing_dict.items())
+
+    def reset(self):
+        """Clear the dictionary
+        """
+        self.timing_dict = {}
```
ctslater: `self._timing_dict` for a "private" variable (to the extent python lets things be private).
I think having a `reset()` is a little confusing, and it seems like it would be more manageable to just create a new instance.

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -381,6 +392,27 @@ def _fillChunkDir(self, chunk_id, neighbor_chunks):
...
+
+        self._data_gen.timingdict.reset()
+        tables = self._data_gen.make_chunk(chunk_id, num_rows=row_counts, seed=self._seed,
+                                    edge_width=self._edge_width, edge_only=edge_only)
+        self._data_gen.timingdict.increment()
+        self._timing_dict.combine(self._data_gen.timingdict)
+        print("tables=", tables)
+
+        for table_name, table in tables.items():
+            edgeType = "CT"  # complete
```
ctslater: `edgetype = "EO" if edge_only else "CT"`

ctslater (commented):
Mostly python comments, overall structure looks good.

python/lsst/dax/data_generator/timingdict.py
jgates108: It's used by the server when importing the dictionaries from the clients. I agree about the name though, maybe self._times?

python/lsst/dax/data_generator/timingdict.py
jgates108: Mostly, it's easier to transmit from client to server. It may be useful to expand the concept to have different counts like "count_neigh" and have sub elements "_neigh_{whatever}" for different averages. Thoughts?

ctslater: Hmm the root of this problem really lies with doing custom serialization, because now it either pushes complexity into that serialization to handle details of the objects, or it makes the objects more complex to use so they can be simpler to serialize. Can pickling be used instead? It would make it a lot easier if it could at least handle nested dictionaries. At the very least, I'd make the serialization code responsible for separating out "count" keys from the timing keys instead of leaving them in the same dictionary. I don't really understand the relationship between what is being counted vs. the timings stored in different keys, so I can't evaluate proposals re: "count_neigh" etc.

I know this is just a tiny timing class and I'd hate for it to take a ton of effort, but I can easily imagine wanting to extend the network communication further and we'll just run into the same problems again.

python/lsst/dax/data_generator/timingdict.py
ctslater: The main issue I have is that if `__init__()` just sets the instance variable to the parameter, that breaks encapsulation and makes it harder to do something like using a `defaultdict` internally. `combine()` is doing the right thing by taking the values, not the object itself, so it might be better to either let the server create an empty instance and call combine, or call combine from inside `__init__()`.

`self._timing_dict` I think is fine inside the class, it's just when it appears outside or in the function signature that it's confusing.

python/lsst/dax/data_generator/timingdict.py
jgates108: Pickling sounds like a good idea.

python/lsst/dax/data_generator/timingdict.py
jgates108: I need the loop for the sum of values anyway. Is doing separate loops more efficient?

ctslater: It's not faster but the difference is microseconds. The main reason I recommend it is that saying `width =` at the top level makes it really clear what the output is

python/lsst/dax/data_generator/timingdict.py
jgates108: Pickling has issues with data formats and it looks like it is really mean to write things to disk. count is now its own member variable.

python/lsst/dax/distribution/DataGenConnection.py
jgates108: Making count a member changes this. I went to 
`for j in range(1, len(msg_split), 2):
            timing_dict.add(msg_split[j], msg_split[j+1])`

Adding another separator would be worse, IMO.

ctslater: Sounds good to me. I'm guessing `range` starts at 1 because the first element is count, but don't you need `len(msg_split) - 1`?

python/lsst/dax/distribution/DataGenConnection.py
jgates108: This gave me trouble. Single character separators wont work when transmitting generated strings, as things like ';' or ',' are guaranteed to show up eventually, and the join is causing issues when the dictionary is empty.

ctslater: Both ; and , can be full strings (they can be your `sep` and `c_sep` variables), no reason they have to be characters. I think it's the `[:50]` slice that's causing trouble when the dictionary is empty. It works if you make it `list(timing_dict.items())[:50]`, but I'll admit that's getting rather convoluted.

python/lsst/dax/data_generator/timingdict.py
jgates108: I get why people like constructs like `width = max(len(x) for x in self.timing_dict.keys())` but they keep failing for empty sequences. Is there a way to deal with that without checking the list isn't empty first?

python/lsst/dax/distribution/DataGenConnection.py
jgates108: I took the [:50] out and was using c_sep in place of ';' and ',' and it still balked on an empty dictionary.

python/lsst/dax/data_generator/generator.py
```diff
@@ -55,6 +56,7 @@ def __init__(self, spec):
 
         self.spec = spec
         self.tables = spec.keys()
```
kennywlo, line 58: It'll be good to provide some descriptions of what spec and spec.keys are in the constructor.  Also, consider just keeping spec only.

python/lsst/dax/data_generator/timingdict.py
jgates108: `width = 5 if not self.times else max(len(x) for x in self.times.keys())`

python/lsst/dax/data_generator/generator.py
```diff
@@ -55,6 +56,7 @@ def __init__(self, spec):
...
+        self.timingdict = TimingDict()
 
     @staticmethod
     def _resolve_table_order(spec):
```
kennywlo, line 62: Just don't see why this internal method has to be a static method, as spec being passed in is self.spec.

python/lsst/dax/data_generator/generator.py
```diff
@@ -191,6 +196,7 @@ def make_chunk(self, chunk_id, num_rows=None, seed=1, edge_width=0.017, edge_onl
                             prereq_row=preq_rw,
                             prereq_tables=prereq_table_contents)
                         self._add_to_list(col_info.block, output_columns, split_column_names)
+                self.timingdict.end(f'gen_{table}_{col_info.col_names}', st_time)
```
kennywlo: Use double quotes here, just for consistency?

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -26,9 +26,12 @@
 import shutil
 import socket
 import subprocess
+import time
```
kennywlo: Unused import.

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -208,6 +217,7 @@ def _readDatagenConfig(self):
         self._edge_width = spec_globals['edge_width']
         print("_cfgFileContents=", self._cfg_file_contents)
         print("_spec=", self._spec)
+        self._data_gen = DataGenerator(self._spec)
```
kennywlo, line 223: Would like to see a comment on what and why of exec() statement above.

python/lsst/dax/distribution/DataGenConnection.py
```diff
@@ -21,12 +21,15 @@
 
 import socket
```
kennywlo: Unused import.

python/lsst/dax/distribution/DataGenConnection.py
```diff
@@ -48,6 +51,7 @@ class DataGenConnection():
     S_PCFG_A = 'S_PCFG_A' # server answering with specified config file
     C_CHUNKR = 'C_CHUNKR' # client request for chunks to generate
     S_CNKLST = 'S_CNKLST' # server sending chunks to generate
+    C_TIMDCT = 'C_TIMDCT' # client sending timing information
     C_CKCOMP = 'C_CKCOMP' # client sending list of chunks completed
     C_CKCFIN = 'C_CKCFIN' # marks the end of the list
```
kennywlo: Should these variables all be prefixed with _, if only used within the class?

python/lsst/dax/distribution/DataGenConnection.py
```diff
@@ -418,6 +429,47 @@ def clientRecvChunks(self):
...
+        Return
+        ------
+        timing_dict : TimingDict
+            TimingDict object
+        """
+        print("servRecvTiming C_TIMDCT")
+        msg_id, msg, msg_len = self._recv_msg()
+        if msg_id != self.C_TIMDCT:
+            self.warnings += 1
+            raise DataGenError(f'ERROR servRecvTiming {str(msg_id)}:{str(msg_len)} {msg}')
```
kennywlo, line 460: Why the warnings increment if it's not being returned in the raise()?

jgates108: At the end of running tests, warnings should be a specific number. The exception may have been caught and gone un-noticed.

python/lsst/dax/distribution/testingdistrib.py
```diff
@@ -59,7 +63,7 @@ def run(self):
             # receive init from client
             serv.servReqInit()
             # server sending back configuration information for datagenerator
-            serv.servRespInit(self.name, self.arg_string, self.cfg_file_contents, self.ingest_dict)
+            serv.servRespInit(self.name, self.objects, self.visits, self.seed, self.cfg_file_contents, self.ingest_dict)
```
kennywlo: Line is too long (>120).

python/lsst/dax/distribution/testingdistrib.py
```diff
@@ -100,35 +111,41 @@ class ClientTestThrd(threading.Thread):
...
             client = DataGenConnection.DataGenConnection(s)
             client.clientReqInit()
-            name, arg_string, cfg_file_contents, ingest_dict = client.clientRespInit()
+            name, objects, visits, seed, cfg_file_contents, ingest_dict = client.clientRespInit()
             print("ingest_dict=", ingest_dict)
-            if (name == self.name and arg_string == self.arg_string
+            if (name == self.name
+                and objects == self.objects
+                and visits == self.visits
+                and seed == self.seed
```
kennywlo, line 144: An explanation on the long and statement here would help to understand why it works.

python/lsst/dax/data_generator/generator.py
jgates108: I don't know, but this is Colin's code and he may have his reasons. I would guess it makes for easier testing, experimenting.

python/lsst/dax/data_generator/timingdict.py
```diff
@@ -0,0 +1,136 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import time
+
+
+class TimingDict:
```
kennywlo, line 25: Why is TimingDict not based off dict despite its name?

python/lsst/dax/distribution/DataGenClient.py
jgates108: I added some brief comments. 'spec' itself is very complicated in what can be done with it. I expect Colin will be making significant changes to how it works and he will, hopefully, write something on how it should be used. See configs/fakedb/fakeGenSpec.py for an example.

kennywlo (commented):
Looking pretty good overall.

python/lsst/dax/distribution/testingdistrib.py
jgates108: I added a comment. If the values sent from the server thread across the connection don't match what is expected, there's a problem with encoding/decoding the messages.

python/lsst/dax/data_generator/timingdict.py
jgates108: It contains a dictionary of times and I really don't want to override the basic dictionary methods.

kennywlo (commented):
This project is missing the LICENSE and COPYRIGHT statements.

PR #4, opened 2020-09-11
Tickets/dm 26561

jgates108:


python/lsst/dax/distribution/chunklogs.py
```diff
@@ -0,0 +1,412 @@
...
+        if self._target_raw:
+            raw_in = ChunkListFile(None)
+            # Use comma for separator since this came from the command line.
+            raw_in.parse(self._target_raw, ',')
+
+        if self._target._fname:
+            # Use the file, if provided. If raw text input
+            # was provided, use the intersection of that and the
+            # target file.
+            self._target.read()
```
kennywlo, line 235: May help to add a comment on what this read() does.

python/lsst/dax/distribution/chunklistfile.py
```diff
@@ -9,30 +26,259 @@ class ChunkListFile:
...
+        target_raw : str
+            If not None, create _target from this string.
+
+        Note
+        ------
+        self.result_set is produced here. it is a set of target chunks with
+        all invalid, completed, assigned, and limbo chunks removed.
+        """
+
+        if self._target._fname:
```
kennywlo: Accessing protected member _fname of _target.

python/lsst/dax/distribution/DataGenServer.py
```diff
@@ -195,25 +190,37 @@ def __init__(self, cfg_file_name, min_chunk_num, max_chunk_num,
...
 
         # Track all client connections so it is possible to
         # determine when the server's job is finished.
         self._active_client_count = 0
         self._active_client_mtx = threading.Lock()
 
+    def chunksToSendTotal(self):
+        """Return the total number of chunks to send.
+        """
+        return self._chunks_to_send_total
```
kennywlo, line 220: Why not make _chunks_to_send_total public variable instead of adding this accessor method?

python/lsst/dax/distribution/chunklogs.py
```diff
@@ -0,0 +1,413 @@
...
+        # this could just be target - assigned, but users
+        # are expected to edit the files and ingesting the
+        # same chunk twice could be bad.
+        self.result_set = self._target.chunk_set.copy()
+
+        # Remove chunks from result_set if they are in completed,
+        # assigned, or limbo.
+        lst = [self._completed, self._assigned, self._limbo]
+        for item in lst:
+            if item._fname:
```
kennywlo, line 259: Isn't _fname supposed to be private?

python/lsst/dax/distribution/DataGenServer.py
```diff
@@ -474,14 +487,18 @@ def start(self):
...
         print("Done, generated ", self._total_generated_chunks)
+
         print("chunks failed chunks:", self.chunksInState([GenerationStage.LIMBO, GenerationStage.ASSIGNED]))
         counts = {GenerationStage.UNASSIGNED:0,
-            GenerationStage.ASSIGNED:0,
-            GenerationStage.FINISHED:0,
-            GenerationStage.LIMBO:0}
+                  GenerationStage.ASSIGNED:0,
+                  GenerationStage.FINISHED:0,
+                  GenerationStage.LIMBO:0}
```
kennywlo, line 493: What's the advantage of using enum here vs a simple Dict for counts?

python/lsst/dax/distribution/chunklogs.py
```diff
@@ -0,0 +1,413 @@
...
+                    val_a = int(st_split[0])
+                    val_b = int(st_split[1])
+                    if val_a > val_b:
+                        tmp = val_a
+                        val_a = val_b
+                        val_b = tmp
+                    for j in range(val_a, val_b + 1):
+                        self.chunk_set.add(j)
+                else:
+                    raise ValueError(f"value error in st={st} {st_split}")
```
kennywlo, line 63: Can we try to reduce the nesting here?  For example, check != 2 for ValueError() first, then rest can be flattened one level up.

python/lsst/dax/distribution/chunklogs.py
```diff
@@ -0,0 +1,413 @@
...
+        """
+        if path_header is None:
+            path_header = ''
+        targf, compf, assif, limbf = ChunkLogs.createNames(path_header)
+        logs_out = ChunkLogs(targf, compf, assif, limbf)
+        logs_out._target.chunk_set = self._target.chunk_set.copy()
+        logs_out._completed.chunk_set = self._completed.chunk_set.copy()
+        logs_out._assigned.chunk_set = self._assigned.chunk_set.copy()
+        logs_out._limbo.chunk_set = self._limbo.chunk_set.copy()
+        logs_out.result_set = self.result_set.copy()
```
kennywlo, line 327: In regard to Python convention: _target, _completed, _assigned, etc, should not be prefixed with _ if they're to be accessed outside their class.

kennywlo (commented):
Looking good.  Just a few comments on documentation and coding style.

python/lsst/dax/distribution/chunklogs.py
jgates108: logs_out and self are both ChunkLogs objects, so they are allowed to access each others private members.

python/lsst/dax/distribution/chunklogs.py
jgates108: In C++ I would have made ChunkLogs a friend of ChunkListFile, as they are closely related classes. I don't really want anything else (at least outside this file) looking at _fname. I'm not sure how to set that up with python.

python/lsst/dax/distribution/DataGenServer.py
jgates108: I want it to be impossible for some other class to change its value.

python/lsst/dax/distribution/DataGenServer.py
jgates108: Mostly, it gives runtime errors if one of the values is misspelled.

PR #5, opened 2020-09-14
DM-26529: Add density model to dax_data_generator

ctslater:


python/lsst/dax/data_generator/generator.py
```diff
@@ -137,73 +137,113 @@ def make_chunk(self, chunk_id, num_rows=None, seed=1, edge_width=0.017, edge_onl
...
-                        # Reducing length to what was generated for RA and Dec
-                        table_length = blockLength
-                    self.timingdict.end("gen_raDecGen", st_time)
-                cols.append(col_info)
-            table_columns[table] = cols
-            rows_per_table[table] = table_length
-
-        for table in resolved_order:
-            cols = table_columns[table]
+            #cols = table_columns[table]
```
jgates108: This probably can be removed.

python/lsst/dax/data_generator/generator.py
```diff
@@ -137,73 +137,113 @@ def make_chunk(self, chunk_id, num_rows=None, seed=1, edge_width=0.017, edge_onl
...
+            else:
+                density_model = self.spec[table]["density"]
+                chunk_latlon = self.chunker.getChunkBounds(chunk_id).getCenter()
+                ra_center = chunk_latlon.getLon().asDegrees()
+                dec_center = chunk_latlon.getLat().asDegrees()
+                chunk_density = density_model.get_density_at_point(ra_center, dec_center)
+
+            generated_data_per_box = []
+            boxes = self._make_subchunk_boxes(chunk_id, edge_width=edge_width,
+                                                                  edge_only=edge_only)
```
jgates108: alignment

python/lsst/dax/data_generator/generator.py
```diff
@@ -137,73 +137,113 @@ def make_chunk(self, chunk_id, num_rows=None, seed=1, edge_width=0.017, edge_onl
...
+                    temp = np.concatenate(output[name])
+                    output[name] = temp
+                generated_data_per_box.append(pd.DataFrame(output))
+
+
+            output_tables[table] = pd.concat(generated_data_per_box)
 
         return output_tables
 
+    def _make_subchunk_boxes(self, chunk_id, edge_width=0, edge_only=False):
```
jgates108: Calling these 'subchunk' boxes is likely to confuse people. Maybe just '_make_boxes'?

python/lsst/dax/data_generator/generator.py
```diff
@@ -137,73 +137,113 @@ def make_chunk(self, chunk_id, num_rows=None, seed=1, edge_width=0.017, edge_onl
...
+            # Correct the edge_width for declination so there is at least
+            # edge_width at both the top and bottom of the east and west blocks.
+            edge_raA = edge_width / math.cos(decA + edge_width)
+            edge_raB = edge_width / math.cos(decB - edge_width)
+            edge_widthRA = max(edge_raA, edge_raB)
+
+            box_north = SimpleBox(raA, raB, decB - edge_width, decB)
+            box_east = SimpleBox(raA, raA + edge_widthRA, decA + edge_width, decB - edge_width)
+            box_west = SimpleBox(raB - edge_widthRA, raB, decA + edge_width, decB - edge_width)
+            box_south =SimpleBox(raA, raB, decA, decA + edge_width)
```
jgates108: space after =

python/lsst/dax/data_generator/generator.py
```diff
@@ -137,73 +137,113 @@ def make_chunk(self, chunk_id, num_rows=None, seed=1, edge_width=0.017, edge_onl
...
+        if ra_delta < 0:
+            raA = raA - 360.0
+            ra_delta = raB -raA
+        decA = chunk_box.getLat().getA().asDegrees()
+        decB = chunk_box.getLat().getB().asDegrees()
+
+        boxes = []
+        print("chunk=", chunk_id, "bbox=", chunk_box.__repr__())
+
+        if edge_width > 0.0:
```
jgates108: I thought there was limit here where if edge_width was really big compared to the size of the chunk, it just made the entire chunk. I don't think good things happen if the the bottom box overlaps the top box. Granted, this shouldn't happen.

tests/test_generator.py
```diff
@@ -3,34 +3,37 @@
...
 
-        seed = 1
         edge_width = 0.017 # degrees
-        data = generator.make_chunk(chunk_id, num_rows=50, seed=seed,
-                                    edge_width=edge_width, edge_only=False)
+        data = generator.make_chunk(chunk_id, edge_width=edge_width, edge_only=False)
+        chunk_area = chunker.getChunkBounds(chunk_id).getArea() / RAD_PER_DEG**2
         self.assertIn('Object', data.keys())
-        self.assertEqual(len(data['Object']), 50)
+        self.assertAlmostEqual(len(data["Object"]), int(50 * chunk_area), delta=1)
```
jgates108: Would a higher density make for a better test by reducing the impact of delta=1?

tests/test_generator.py
```diff
@@ -51,37 +54,39 @@ def testCcdVisit(self):
...
         edge_width = 0.017 # degrees
-        generator = DataGenerator(generator_spec)
-        chunk_table = generator.make_chunk(chunk_id, num_rows=50, seed=seed,
-                                           edge_width=edge_width, edge_only=False)
+        generator = DataGenerator(generator_spec, chunker, seed=seed)
+        chunk_area = chunker.getChunkBounds(chunk_id).getArea() / RAD_PER_DEG**2
+        chunk_table = generator.make_chunk(chunk_id, edge_width=edge_width, edge_only=False)
         self.assertIn("CcdVisit", chunk_table.keys())
-        self.assertEqual(len(chunk_table["CcdVisit"]), 50)
+        self.assertAlmostEqual(len(chunk_table["CcdVisit"]), int(50 * chunk_area), delta=1)
```
jgates108: I'm confused as to why CcdVisit has anything to do with object density.

jgates108 (commented):
I t mostly looks pretty good. bin/edgefirsttest.py failed, and I'd feel better if it passed. 'python bin/datagenserver.py -k' failed in a way that suggest something needs to be done with  configs/fakedb/fakeGenSpec.py.

ctslater:
Thanks @jgates108. Your comments made me realize that it made more sense to put the visit-precomputation work on this ticket, so that this isn't a weird halfway state of generating visits inside each chunk (which made the tests in particular seem really weird). 

I've thus added some new commits that allow tables to be loaded from an existing CSV file, and a script to generate visit  records for the entire survey. This means that every chunk gets the entire visit table, and needs to decide which are overlapping the chunk center. It overall seems pretty quick to do that. The edgefirst tests and the unit tests all pass now.

The other important addition is a new spec file that has the entire set of Object table columns, as specified in LDM-153. I think this is approximately the right size scale for testing.

tests/test_generator.py
ctslater: Right, this is now removed since visits are precomputed.

python/lsst/dax/data_generator/columns.py
```diff
@@ -395,20 +307,60 @@ def __init__(self, n_mags=1, min_mag=0, max_mag=27.5, column_val=7):
...
+    column_val : int
+        Arbitrary integer that should be different from other
+        column values. Used in random number generation.
+
+    Note
+    ----
+    Currently generates a flat magnitude distribution. Should properly
+    be some power law.
+    If there is more than one call to this in a single table, there will
+    be correlation between rows as the same random numbers will be used.
```
jgates108: It looks like you are using column_val to keep this from being an issue. Maybe expand the comment.

python/lsst/dax/data_generator/columns.py
```diff
@@ -462,41 +414,38 @@ def __init__(self, filters="ugrizy", visit_radius=0.30, column_val=3):
         self.visit_radius = visit_radius
         self.column_val = column_val
 
-    def __call__(self, chunk_id, length, seed, prereq_row=None, prereq_tables=None):
-        assert prereq_row is not None, "ForcedSourceGenerator requires rows from Object."
+    def __call__(self, box, length, seed, prereq_row=None, prereq_tables=None, unique_box_id=0,
```
jgates108, line 406: I think the TODO comment above is out of date now.

python/lsst/dax/data_generator/generator.py
```diff
@@ -110,18 +115,13 @@ def _add_to_list(self, generated_data, output_columns, split_column_names):
                 raise ValueError("Column name implies multiple returns, "
                                  "but generator only returned one")
 
-    def make_chunk(self, chunk_id, num_rows=None, seed=1, edge_width=0.017, edge_only=False):
+    def make_chunk(self, chunk_id, edge_width=0.017, edge_only=False, return_pregenerated=False):
```
jgates108, line 118: return_pregenerated isn't in Parameters

tests/test_columns.py
```diff
@@ -82,41 +86,50 @@ def testFSGenerator(self):
...
+        prereq_tables = {'CcdVisit': visit_df, 'Object': object_df}
+
         cell_id = 1
         length = 0
-        fs_generator = columns.ForcedSourceGenerator(visit_radius=0.30)
+        seed = 1
+        fs_generator = columns.ForcedSourceGenerator(visit_radius=1.00)
+
+        # (array([1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 6, 7, 7, 8, 8]), array([101, 102, 101, 102, 101,
+        # 102, 101, 102, 101, 102, 101, 102, 101, 102, 101, 102]), a
```
jgates108: Comment still needed?

jgates108 (commented):
I like it, it's much improved. There are a couple of comments that should be cleaned up.

I do have some questions about how "visit_table_chunk3525.csv" was generated. It isn't much work to pass a particular file from the server to the clients, but it may be better to have the server generate the visits file, or pass the instructions to workers and have them build the visit files. I guess it comes down to how big the visits files are expected to get as sending large files could overwhelm the server. This is probably something for another ticket. As it is, the server doesn't send the visits file to the clients, so I'm not sure if that's a different ticket to fix that or what you would like to do.

ctslater:
So that brings up a bit of a deployment question. In addition to a visits file, a better density model will also need to have some big-ish data file available on the client. Does it make sense to just wrap those up into the docker image, so there's no need to transport them nor parameterize them enough to be generated on the fly?

jgates108:
I'd really rather not put these in the docker files, since that requires new containers to be built for new data sets. That gets awkward. How big really matters. A few MB wouldn't be that bad, a couple of GB probably would be too much.

PR #6, opened 2020-09-25
Tickets/dm 26835

jgates108:


python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -252,18 +261,19 @@ def findCsvInTargetDir(self, chunk_id, neighbor_chunks):
...
+        # TODO: python3 has better version of glob that would make this code cleaner.
         chunkCsvPaths = glob.glob(findCsv)
         print("findCsv=", findCsv, " paths=", chunkCsvPaths)
         # Remove the path and add the files to a list
         chunkCsvFiles = []
         for fn in chunkCsvPaths:
             chunkCsvFiles.append(os.path.basename(fn))
         print("chunkCsvFiles=", chunkCsvFiles)
-        # These cannot be edgeOnly, and there must be one for each entry in self._spec['spec']
+        # These cannot be edgeOnly, and there must btraceback.print_exc()
```
ctslater: Some sort of search and replace error here?

python/lsst/dax/distribution/DataGenClient.py
```diff
@@ -310,7 +320,7 @@ def removeFilesForChunk(self, chunk_id, edge_only=False, complete=False):
         complete : bool
             When True, remove complete csv and parquet files
 
-        Return
+        Returntraceback.print_exc()
```
ctslater: Same search and replace error?

python/lsst/dax/distribution/DataGenClient.py
jgates108: I guess!? That's really odd.

configs/fakedb/ingestCfgs/fakedb_ForcedSource.json
```diff
@@ -0,0 +1,17 @@
...
+    "is_director":0,
+    "director_key":"objectId",
+    "schema":[
+     {"name":"objectId","type":"BIGINT NOT NULL"},
+     {"name":"ccdVisitId","type":"BIGINT NOT NULL"},
+     {"name":"psFlux","type":"DOUBLE"},
+     {"name":"psFlux_Sigma","type":"DOUBLE"},
+     {"name":"chunkId","type":"INT UNSIGNED NOT NULL"},
+     {"name":"subChunkId","type":"INT UNSIGNED NOT NULL"}
+    ],
```
kennywlo, line 15: I'm quite confused about the definition of the "schema" above.  See https://json-schema.org/learn/miscellaneous-examples.html of what I used to see how schemas defined in JSON.

configs/fakedb/partitionerCfgs/ForcedSource.cfg
```diff
@@ -5,24 +5,14 @@
...
 part = {
     # The ID of the object associated with each source -
     # points into the Object table.
     id = objectId
-    # The partitioning position is the position of the associated object.
-    pos = 'raObject, declObject'
     # Omit overlap for sources - this means spatial joins involving the
     # Source table directly will not work.
-    overlap = 0.01667
+    overlap = 0
```
kennywlo: Sure it's not 0.0 for overlap?

python/lsst/dax/distribution/DataGenConnection.py
```diff
@@ -582,7 +588,8 @@ def testMethods(self):
         if len(diff) != 0:
             print("large range chunkMsg failed", diff)
             success = False
-        if success is None: success = True
+        if success is None:
+            success = True
```
kennywlo: That's a special case of success, which not expected to happen often?, so a print message here could help to track its occurrence.

python/lsst/dax/distribution/testingdistrib.py
```diff
@@ -184,9 +186,9 @@ def run(self):
             while len(completedChunks) > 0:
                 completedChunks = client.clientReportChunksComplete(completedChunks)
 
-
         print("ClientTestThrd.run finished")
-        if self.success is None: self.success = True
+        if self.success is None:
+            self.success = True
```
kennywlo, line 191: Same as above, print message to track this special case of success?

kennywlo (commented):
Good overall.  A few comments on styling.

configs/fakedb/ingestCfgs/fakedb_ForcedSource.json
jgates108: It would be nice if it was db_schema or something to separate it from json schema, but these are configuration files for the ingest system, so I can't really change them.

python/lsst/dax/distribution/DataGenConnection.py
jgates108: This isn't a special occurrence, this is just me being uptight about saying something is successful before it is actually successful.

python/lsst/dax/distribution/testingdistrib.py
jgates108: This one actually serves a purpose. Since success = None until the job is complete, other entities can test success and see that the job isn't done, and if an exception is raised, this line is never called and success remains not True.

kennywlo (approved):
All looking good to merge.

PR #8, opened 2020-10-14
Tickets/dm 27087

jgates108:


bin/makevisits.py
```diff
@@ -21,15 +21,46 @@
...
+    print('-n, --numOfVisits Number of visits')
 
-if __name__ == '__main__':
 
-    # 1000 visits per night, 0.5 years of observing.
-    n_visits = 1000*150
-    area = 20000
-    north_only = True
+def create_visits():
+    """Start the server.
```
ctslater, line 40: Copy/paste error?

python/lsst/dax/data_generator/columns.py
```diff
@@ -418,19 +419,27 @@ def __call__(self, box, length, seed, prereq_row=None, prereq_tables=None, uniqu
...
+        min_dec = box.decA - v_radius
+        max_dec = box.decB + v_radius
+        if (edge_only):
+            trimmed_visit = pd.DataFrame(data=None, columns=visit_table.columns)
+        else:
+            trimmed_visit = visit_table.loc[(visit_table['decl'] >= min_dec) &
+                                            (visit_table['decl'] <= max_dec)]
+        print(f"edge_only={edge_only} len trimmed={len(trimmed_visit)}  base={len(visit_table)}")
+
+        visit_skycoords = SkyCoord(ra=trimmed_visit['ra'], dec=trimmed_visit['decl'], unit="deg")
```
ctslater, line 456: What's the actual speed up on this in terms of overall runtime?  I'd be hesitant about sacrificing generality for this unless it's really where most of the time is going.

python/lsst/dax/data_generator/generator.py
```diff
@@ -169,27 +172,30 @@ def make_chunk(self, chunk_id, edge_width=0.017, edge_only=False, return_pregene
...
                 box_rowcount = int(chunk_density * box.area())
                 if(box_rowcount == 0):
                     continue
                 unique_box_id = chunk_id*8 + box_n
+                box_center_ra = (box.raA + box.raB)/2.0
+                box_center_dec = (box.decA + box.decB)/2.0
+                box_center = SkyCoord(box_center_ra, box_center_dec, frame="icrs", unit="deg")
                 output = self._generate_table_block(box, column_generators,
-                                                    chunk_center=chunk_center,
+                                                    chunk_center=box_center,
```
ctslater: I had made this the chunk center so that the effective visit boundaries would be at least somewhat understandable when looking at the resulting data (one would see chunk boundaries, but wouldn't need to know anything about overlap areas). Is there a functional reason to make this the box center? The parameter should at least be renamed if that's the case.

python/lsst/dax/data_generator/columns.py
```diff
@@ -418,19 +419,27 @@ def __call__(self, box, length, seed, prereq_row=None, prereq_tables=None, uniqu
                                           (object_table['decl'] >= box.decA) &
                                           (object_table['decl'] < box.decB)]
 
-        visit_skycoords = SkyCoord(ra=visit_table['ra'], dec=visit_table['decl'], unit="deg")
+        v_radius = self.visit_radius * 1.5  # Go a little bit bigger so nothing is missed.
+        min_dec = box.decA - v_radius
+        max_dec = box.decB + v_radius
+        if (edge_only):
```
ctslater, line 449: Should put a comment here explaining that this is an intentional policy choice not to make FS in this case. I couldn't figure it out until I re-read the slack discussion.

configs/kpm50/serverCfg.yml
```diff
@@ -0,0 +1,27 @@
+# serverCfg.yml
+server:
+  port: 13042
+
+fakeDataGenerator:
+# --chunk 3525 --visits 30 --objects 10000 example_spec.py
+  arguments: '--visits 30 --objects 1000'
```
ctslater: I'm guessing these arguments are outdated? But I'm not sure.

python/lsst/dax/data_generator/columns.py
jgates108: It's about 100 times faster. (From 2 hours to less than 5 minutes for valid chunks from 0 to 2000.)

python/lsst/dax/data_generator/generator.py
jgates108: I ran into an issue with the trimming process above missing some of the points in the north and south boxes due to the chunk center being different than the box center. I can change the name.

ctslater: It looks like that might be because you do the trimming based on the box center, but the actual calculation was done on the chunk center. You have access to the chunk center in the function doing the trimming, wouldn't using that solve the inconsistency?

jgates108: The function really operates on boxes, so sticking with box boundaries and box centers seems more appropriate.

PR #9, opened 2020-10-26
DM-27263: Generate ingest and partitioner configs from yaml

ctslater:


configs/kpm50/fakeGenSpec.py
```diff
@@ -10,37 +10,54 @@
...
+            "objectId,ccdVisitId,psFlux,psFlux_Err,flags":
                 columns.ForcedSourceGenerator(visit_radius=1.7, filters="ugriz"),
         },
     }
 }
+
+spec["CcdVisit"] = {
+    "from_file": "visit_table_chunk3525.csv",
+    "columns": "ccdVisitId,filterName,ra,decl"
+}
```
jgates108: Why is this isolated from the rest of the definition of 'spec'?

configs/makeIngestSpec.py
```diff
@@ -0,0 +1,136 @@
...
+
+    output_field = output_json["out"]["csv"]["field"]
+    for n in range(len(output_field)):
+        if(len(output_field[n]) == 0):
+            output_field[n:n+1] = schema_columns
+
+    input_field = output_json["in"]["csv"]["field"]
+    for n in range(len(input_field)):
+        if(len(input_field[n]) == 0):
+            input_field[n:n+1] = schema_columns
```
jgates108, line 82: I think it may be a bit better if looked for an entry like {"name":"","type":"SCHEMA_PLACEHOLDER"} instead of a zero length field. I don't think it makes that much difference, but being explicit may be a bit clearer what is happening in the json file.  Same applies to transform_ingest_json.

jgates108 (approved):
Looks good, just a suggestion in the comments.

PR #10, opened 2020-12-09
Tickets/dm 27472

jgates108:
The major change was moving transaction start and end to the server instead of on the clients.

There is something happening where changing the number of chunks in a transaction changes the number of objects in the Object table. I'm creating another ticket to deal with that, but if you see a likely cause, feel free to point it out.

A significant amount of code was moved from DataGenServer.py to chunktracking.py. ChunkTracking controls and tracks which chunks are part of each transaction and closes each transaction when all of the chunks for that transaction have been reported ingested. More than one transaction can be open at a time.


configs/fakedb/README
```diff
@@ -11,3 +11,7 @@ To build it, skipping data ingest with the replicator:
   cd dax_data_generator
   bin/run_distribserver.sh fakedb -glocalhost -k -r "0:2000"
 
+
+Make sure stripes and substripes are consistent:
+grep -R stripe
+grep -R overlap
```
iagaponenko: No newline in the end of the file.

configs/kpm50/README
```diff
@@ -4,3 +4,7 @@ To build it:
   Set the system up according to dax_data_generator/README.md, then:
     cd dax_data_generator
     bin/run_distribserver.sh kpm50 -glocalhost -k -r "0:2000"
+
+Make sure stripes and substripes are consistent:
+grep -R stripe
+grep -R overlap
```
iagaponenko: No newline in the end of the file.

iagaponenko (approved):
I have looked at the changes. Quite frankly, it's hard for me to make any statements on its correctness since this is an incremental development on top of some complex software of which I don't know much, except it's intents and general functionality. So, I leave up to someone else to do this.

On my side, I would make just one comment on the chunk-to-worker placement algorithm (the one calling Master Controller's service `/ingest/chunks`). The latest version of the Master Controller allows using the name of a database to specify a context for the chunks to be allocated/placed:
```
{"database":<string>,
 "chunks":[<number>,<number>,...<number>],
 "auth_key":<string>
}
```
Details can be found at: https://confluence.lsstcorp.org/pages/viewpage.action?pageId=135955129

This may help you to further improve your ingest workflow since you no longer need to start any transactions before allocating chunks. All you need to know is the name of a database.

python/lsst/dax/data_generator/columns.py
```diff
@@ -491,7 +499,7 @@ def __call__(self, box, length, seed, prereq_row=None, prereq_tables=None, uniqu
         psFlux = (np.repeat(objects_inside_box['gPsFlux'].values, n_matching_visits) +
                   np.random.randn(n_rows_total))
         psFluxSigma = np.zeros(n_rows_total) + 0.1
-        flags = np.random.randint(200, size=n_rows_total)
+        flags = np.random.randint(-128, high=127, size=n_rows_total)
```
ctslater: There should never be negative flags, so I'd make this between 0 and 127.

PR #12, opened 2021-01-08
Tickets/dm 27937

jgates108:


python/lsst/dax/data_generator/columns.py
```diff
@@ -517,3 +514,150 @@ def __call__(self, box, length, seed, prereq_row=None, prereq_tables=None, uniqu
...
+        min_dec = box.decA - v_radius
+        max_dec = box.decB + v_radius
+        trimmed_visit = visit_table.loc[(visit_table['decl'] >= min_dec) & (visit_table['decl'] <= max_dec)]
+        print(f"edge_only={edge_only} len trimmed={len(trimmed_visit)}  base={len(visit_table)}")
+
+        # Unlike the ForcedSource table, only some of the objects should have visits, not
+        # all of them. It would be nice to have the brightest objects get visits, but that
+        # isn't that important right now. Just trim objects from objects_inside_box until
+        # the right ratio is reached for now.
+        target_percentage = 0.3  # supplied by Colin
```
ctslater: Minor thing, but I'd make this an instance variable and have it as a parameter in `__init__()`.

python/lsst/dax/data_generator/columns.py
```diff
@@ -517,3 +514,150 @@ def __call__(self, box, length, seed, prereq_row=None, prereq_tables=None, uniqu
...
+        self.column_seed = column_seed
+
+    def random_col(self, min_val, max_val, length, integer=False):
+        if (min_val > max_val):
+            tmp = min_val
+            min_val = max_val
+            max_val = min_val
+        delta_value = max_val - min_val
+        col = delta_value * np.random.rand(length) + min_val
+        if integer:
```
ctslater, line 546:
```
if integer:
    col = col.astype(int)
````

File: python/lsst/dax/data_generator/columns.py [hunk 3]
```diff
@@ -517,3 +514,150 @@ def __call__(self, box, length, seed, prereq_row=None, prereq_tables=None, uniqu
...
+
+    """
+
+    def __init__(self, filters="ugrizy", visit_radius=0.30, column_seed=27):
+        self.filters = filters
+        self.visit_radius = visit_radius
+        self.column_seed = column_seed
+
+    def random_col(self, min_val, max_val, length, integer=False):
+        if (min_val > max_val):
```
ctslater, line 542:
````
if(min_val > max_val):
    min_val, max_val = (max_val, min_val)
````
Not just for style; the current code has a bug.

File: configs/kpm50/fakeGenSpec.py [hunk 4]
```diff
@@ -72,6 +72,13 @@
                 columns.ForcedSourceGenerator(visit_radius=1.7, filters="ugriz"),
         },
     },
+    "Source": {
+        "prereq_tables": ["CcdVisit", "Object"],
+        "columns": {
+            "sourceId:SID,ccdVisitId:VID,filterName:CHAR(1),objectId:OID,ssObjectId:BIGINT,parentSourceId:BIGINT,htmId20:BIGINT,sky:FLOAT,skyErr:FLOATE,psFlux:FLOAT,psFluxErr:FLOATE,psX:FLOAT,psXErr:FLOATE,psY:FLOAT,psYErr:FLOATE,psFlux_psX_Cov:FLOAT,psFlux_psY_Cov:FLOAT,psX_Y_Cov:FLOAT,psLnL:FLOAT,psChi2:FLOAT,psN:INT,psRa:RA,psRaErr:FLOATE,psDecl:DEC,psDeclErr:FLOATE,psFlux_psRa_Cov:FLOAT,psFlux_psDecl_Cov:FLOAT,x:FLOAT,xErr:FLOATE,y:FLOAT,yErr:FLOATE,x_y_Cov:FLOAT,ra:RA,raErr:FLOATE,decl:DEC,declErr:FLOATE,ra_decl_Cov:FLOAT,E1:FLOAT,E1Err:FLOATE,E2:FLOAT,E2Err:FLOATE,E1_E2_Cov:FLOAT,mSum:FLOAT,mSumErr:FLOATE,m4:FLOAT,apN:TINYINT,flags:BIGINT":
```
ctslater: I understand that it's sort of a sunk cost at this point, but I think it's a little cleaner conceptually to do something like
```
"sourceId,ccdVisitId,filterName,objectId,ssObjectId,parentSourceId": 
    columns.SourceGenerator(visit_radius=1.7, filters="ugriz",
                            dtypes="SID,VID,CHAR(1),OID,BIGINT,BIGINT")
```
That would make it so the generic generator code wouldn't need to have the dtype information plumbed through to every `__call__()` method, and only the SourceGenerator would need to take dtypes as a parameter.

also fwiw it's legal syntax to do 
````
"columns": {
    "sourceId:SID,ccdVisitId:VID,filterName:CHAR(1),"
    "objectId:OID,ssObjectId:BIGINT,parentSourceId:BIGINT": columns.SourceGenerator(...),
````

File: python/lsst/dax/data_generator/columns.py [hunk 5]
```diff
@@ -517,3 +514,150 @@ def __call__(self, box, length, seed, prereq_row=None, prereq_tables=None, uniqu
...
+                # Use the list of object RAs
+                # TODO: add a small random component.
+                data = out_obj_ras
+            elif ctype == 'DEC':
+                # Use the list of object Decs
+                # TODO: add a small random component.
+                data = out_obj_decs
+            elif ctype == 'CHAR(1)':
+                # Random filter
+                filters="ugrizy"
```
ctslater: Might as well use `self.filters` since it's already available.

python/lsst/dax/data_generator/columns.py
```diff
@@ -517,3 +514,150 @@ def __call__(self, box, length, seed, prereq_row=None, prereq_tables=None, uniqu
...
+        max_dec = box.decB + v_radius
+        trimmed_visit = visit_table.loc[(visit_table['decl'] >= min_dec) & (visit_table['decl'] <= max_dec)]
+        print(f"edge_only={edge_only} len trimmed={len(trimmed_visit)}  base={len(visit_table)}")
+
+        # Unlike the ForcedSource table, only some of the objects should have visits, not
+        # all of them. It would be nice to have the brightest objects get visits, but that
+        # isn't that important right now. Just trim objects from objects_inside_box until
+        # the right ratio is reached for now.
+        target_percentage = 0.3  # supplied by Colin
+        df = objects_inside_box.copy()
```
ctslater: Unused?

python/lsst/dax/data_generator/columns.py
```diff
@@ -517,3 +514,150 @@ def __call__(self, box, length, seed, prereq_row=None, prereq_tables=None, uniqu
...
+            indx += 1
+
+        n_rows_total = n_matching_visits * len(objects_inside_box)
+        length = n_rows_total
+
+        assert len(out_objectIds) == n_rows_total
+        assert len(out_ccdVisitIds) == n_rows_total
+
+        for j in range(len(col_dict)):
+            col_info = col_dict[j]
```
ctslater:
```
for j, col_info in enumerate(col_dict.values())):
```
or drop the `.values()` part if you make it a list instead of a dict.

python/lsst/dax/data_generator/columns.py
```diff
@@ -517,3 +514,150 @@ def __call__(self, box, length, seed, prereq_row=None, prereq_tables=None, uniqu
...
+        spec_c = spec_cols.split(",")
+        col_dict = {}
+        indx = 0
+        for cn in spec_c:
+            col_info = {}
+            cname, ctype = cn.split(":")
+            col_info['cname'] = cname
+            col_info['ctype'] = ctype
+            col_info['data'] = None
+            col_dict[indx] = col_info
```
ctslater: `col_dict` is effectively just a list, so this can be
````
col_list = []
for cn in spec_c:
    cname, ctype = cn.split(":")
    col_list.append({"cname": cname, "ctype": ctype, "data": None})
````

File: python/lsst/dax/data_generator/columns.py [hunk 9]
```diff
@@ -517,3 +514,150 @@ def __call__(self, box, length, seed, prereq_row=None, prereq_tables=None, uniqu
...
+        object_table = prereq_tables['Object']
+
+        objects_inside_box = object_table[(object_table['psRa'] >= box.raA) &
+                                          (object_table['psRa'] < box.raB) &
+                                          (object_table['psDecl'] >= box.decA) &
+                                          (object_table['psDecl'] < box.decB)]
+        v_radius = self.visit_radius * 1.5  # Go a little bit bigger so nothing is missed.
+        min_dec = box.decA - v_radius
+        max_dec = box.decB + v_radius
+        trimmed_visit = visit_table.loc[(visit_table['decl'] >= min_dec) & (visit_table['decl'] <= max_dec)]
```
ctslater, line 567: This usage of`.loc[...]` is the same as `visit_table[...]` (but not all usage of `.loc` works that way). A word of caution that this is a view, not a copy, but I don't think that affects anything here.

ctslater (commented):
Looks good, mostly small python changes and one organizational idea.

configs/kpm50/fakeGenSpec.py
jgates108: With the size of the source table, I found it difficult to line up types with columns in separate strings. It was pretty easy to modify makeingestspec.py to produce the above string from the huge yaml file and then hunt down those items that needed their types changed.

